PR #1, opened 2024-06-21
feat: improve error message

etiennebacher:
Naive attempt to include the last method called in the error message as it is something really useful when the pipeline is quite long:

```r
> as_polars_df(mtcars)$group_by("cyl")$head(-Inf)
Error:
! Evaluation failed in `$head()`.
Caused by error:
! n should not be negative
Run `rlang::last_trace()` to see where the error occurred.
```

Not sure how it would behave with `LazyFrame` since the last method should be `collect()` and the method before that is not necessarily the one that errors.

### DIFF
--- a/R/utils-wrap.R
+++ b/R/utils-wrap.R
@@ -2,7 +2,12 @@ wrap <- function(x, ..., call = parent.frame()) {
   try_fetch(
     x,
     error = function(cnd) {
-      abort("Evaluation failed.", call = call, parent = cnd)
+      err_call <- rlang::error_call(call)[[1]]
+      abort(
+        paste0("Evaluation failed in `$", err_call[[length(err_call)]], "()`."),
+        call = call,
+        parent = cnd
+      )
     }
   )
 


eitsupi:
Thanks for the improvement!

> Not sure how it would behave with `LazyFrame` since the last method should be `collect()` and the method before that is not necessarily the one that errors.

It appears to be as follows.

```r
> pl$DataFrame(a = "a")$lazy()$select(pl$col("a")$cast(pl$Int8))$collect()
Error:
Caused by error:
! ComputeError(ErrString("conversion from `str` to `i8` failed in column 'a' for 1 out of 1 values: [\"a\"]"))
Run `rlang::last_trace()` to see where the error occurred.

> rlang::last_trace()
<error/rlang_error>
Error:
Caused by error:
! ComputeError(ErrString("conversion from `str` to `i8` failed in column 'a' for 1 out of 1 values: [\"a\"]"))
---
Backtrace:
     ▆
  1. └─pl$DataFrame(a = "a")$lazy()$select(pl$col("a")$cast(pl$Int8))$collect()
  2.   ├─neopolars:::wrap(self$`_ldf`$collect()) at neo-r-polars/R/lazyframe-frame.R:34:3
  3.   │ └─rlang::try_fetch(...) at neo-r-polars/R/utils-wrap.R:2:3
  4.   │   ├─base::tryCatch(...)
  5.   │   │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
  6.   │   │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
  7.   │   │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
  8.   │   └─base::withCallingHandlers(...)
  9.   └─self$`_ldf`$collect() at neo-r-polars/R/utils-wrap.R:2:3
 10.     └─neopolars:::.savvy_wrap_PlRDataFrame(...) at neo-r-polars/R/000-wrappers.R:400:5
Run rlang::last_trace(drop = FALSE) to see 4 hidden frames.
```

In current r-polars:

```r
> pl$DataFrame(a = "a")$lazy()$select(pl$col("a")$cast(pl$Int8))$collect()
Error: Execution halted with the following contexts
   0: In R: in $collect():
   0: During function call [pl$DataFrame(a = "a")$lazy()$select(pl$col("a")$cast(pl$Int8))$collect()]
   1: Encountered the following error in Rust-Polars:
        conversion from `str` to `i8` failed in column 'a' for 1 out of 1 values: ["a"]
```

Perhaps the Rust Polars-derived error can be corrected by changing the following part and changing the way the string is formatted.

https://github.com/eitsupi/neo-r-polars/blob/c83da50ecdd60c7a33807fcc39d7032818420204/src/rust/src/error.rs#L6-L28

PR #7, opened 2024-10-10
feat: str namespace for expr

etiennebacher:


R/expr-string.R
```diff
@@ -0,0 +1,1136 @@
...
+    dtype,
+    format = NULL,
+    ...,
+    strict = TRUE,
+    exact = TRUE,
+    cache = TRUE,
+    ambiguous = "raise") {
+  if (!is_polars_dtype(dtype)) {
+    abort("`dtype` must be of class `RPolarsDataType`.")
+  }
```
eitsupi, line 107: Please use `check_polars_dtype()` as like this:
https://github.com/eitsupi/neo-r-polars/blob/2672fce59b94c67655ac0ab7a13a1995db4da437/R/expr-expr.R#L265-L272

And, please check dots by `check_dots_empty0()`

eitsupi (changes requested):
Thanks for working on this!

Rewriting tips from current r-polars:

tests/testthat/test-expr-string.R
```diff
@@ -0,0 +1,931 @@
...
+# })
+
+
+test_that("str$replace", {
+  dat <- pl$DataFrame(x = c("123abc", "abc456"))
+  expect_identical(
+    dat$with_columns(pl$col("x")$str$replace(r"{abc\b}", "ABC")) |>
+      as.list(),
+    list(x = c("123ABC", "abc456"))
+  )
```
eitsupi, line 543: Unlike current r-polars, polars_data_frame can be compared directly with `expect_equal()`, so do not use `as.list()`.
https://github.com/eitsupi/neo-r-polars/blob/d5994dc6a55d0cdea27a77eeb54fa04be701c1fd/tests/testthat/test-as_polars_df.R#L35-L38

src/rust/src/expr/string.rs
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    fn str_replace(&self, pat: &PlRExpr, value: &PlRExpr, literal: bool, n: i32) -> Result<Self> {
+        Ok(self
+            .inner
+            .clone()
+            .str()
+            .replace_n(pat.inner.clone(), value.inner.clone(), literal, n as i64)
+            .into())
+    }
```
eitsupi, line 283: Instead of conversion by `as`, convert from `NumericScalar` via `<Wrap<i64>>::try_from()` as follows.
https://github.com/eitsupi/neo-r-polars/blob/d5994dc6a55d0cdea27a77eeb54fa04be701c1fd/src/rust/src/dataframe/general.rs#L55-L65

Do not use `as.integer()` on the R side because it may cause unexpected conversions.

eitsupi (changes requested):

tests/testthat/test-expr-string.R
etiennebacher: Is `expect_equal()` redefined somewhere? I don't understand how it accepts `polars` objects? Also, it converts i64 values to character, so this breaks in this case (but not if I use `as.list()`):
```
dat <- pl$DataFrame(x = c("110", "101", "010"))
expect_equal(
  dat$with_columns(pl$col("x")$str$to_integer(base = 2)),
  pl$DataFrame(x = c(6, 5, 2))
)

#> `actual$x` is a character vector ('6', '5', '2')
#> `expected$x` is a double vector (6, 5, 2)
```

eitsupi: Hmmm, that is quite odd.

The comparison is defined in the S3 method of `waldo::compare()`.
https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/R/dataframe-s3-waldo.R

eitsupi: > Also, it converts i64 values to character, so this breaks in this case

Sorry, I forgot that was defined here. Could you cast the expected and match the dtype?
https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/R/series-s3-waldo.R#L5

etiennebacher: Thanks, it works like that (example above still fails because original is int and expected is float but that's expected)

src/rust/src/expr/string.rs
```diff
@@ -0,0 +1,390 @@
+use crate::conversion::Wrap;
+use crate::{PlRExpr, RPolarsErr};
```
eitsupi, line 2: We can use `prelude::*` here.

```suggestion
use crate::{prelude::*, PlRExpr, RPolarsErr};
```

https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/src/rust/src/prelude.rs#L1-L3

R/series-s3-waldo.R
```diff
@@ -2,7 +2,7 @@
 compare_proxy.polars_series <- function(x, path) {
   list(
     object = structure(
-      x$to_r_vector(ensure_vector = FALSE, int64 = "character", decimal = "character"),
+      x$to_r_vector(ensure_vector = FALSE, int64 = "integer", decimal = "character"),
```
eitsupi, line 5: Sorry, this wasn't my intention, instead of modifying the contents of `compare`, could you cast the column types of the expected DataFrame to match?
Something like `pl$DataFrame(x = c(6, 5, 2))$cast(x = pl$Int64)`

etiennebacher: Done

etiennebacher:
@eitsupi I'd say that 75% of the string subns is adapted, the rest needs other methods to be implemented first or concerns datetime, which requires more work. I think this can be already merged.

eitsupi (commented):
Thanks for the hard work!

Sorry, but the volume is too large to review right away. Please wait a while.

src/rust/src/expr/string.rs
```diff
@@ -0,0 +1,396 @@
+use crate::{prelude::*, PlRExpr, RPolarsErr};
+// use polars::prelude::StrptimeOptions;
```
eitsupi, line 2: This is binded by `prelude::*`

```suggestion
```

eitsupi (changes requested):

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+
+#' Convert a String column into a Date/Datetime/Time column.
+#'
+#' Similar to the [strptime()] function.
+#'
+#' When parsing a Datetime the column precision will be inferred from the format
+#' string, if given, e.g.: `"%F %T%.3f"` => [`pl$Datetime("ms")`][pl_Datetime].
+#' If no fractional second component is found then the default is `"us"` (microsecond).
+#' @param dtype The data type to convert into. Can be either `pl$Date`,
+#' [`pl$Datetime()`][DataType_Datetime], or `pl$Time`.
```
eitsupi, line 28: dtype page is not added yet.
Please add TODO like:
https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/R/dataframe-frame.R#L1

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' @param format Format to use for conversion. Refer to
+#' [the chrono crate documentation](https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
+#' for the full specification. Example: `"%Y-%m-%d %H:%M:%S"`.
+#' If `NULL` (default), the format is inferred from the data.
+#' Notice that time zone `%Z` is not supported and will just ignore timezones.
+#' Numeric time zones like `%z` or `%:z` are supported.
+#' @param ... Not used.
```
eitsupi, line 35: Please inherit from `rlang::args_dots_empty`
https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/R/expr-datetime.R#L30

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' @param strict If `TRUE` (default), raise an error if a single string cannot
+#' be parsed. If `FALSE`, produce a polars `null`.
+#' @param exact If `TRUE` (default), require an exact format match. If `FALSE`,
+#' allow the format to match anywhere in the target string.
+#' Conversion to the Time type is always exact.
+#' Note that using `exact = FALSE` introduces a performance penalty -
+#' cleaning your data beforehand will almost certainly be more performant.
+#' @param cache Use a cache of unique, converted dates to apply the datetime
+#' conversion.
+#' @inheritParams ExprDT_replace_time_zone
```
eitsupi, line 45: Is this correct?

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+    ...,
+    strict = TRUE,
+    exact = TRUE,
+    cache = TRUE,
+    ambiguous = "raise") {
+  wrap({
+    check_dots_empty0(...)
+    check_polars_dtype(dtype)
+
+    if (pl$same_outer_dt(dtype, pl$Datetime())) {
```
eitsupi, line 109: Does the `pl$same_outer_dt` function exist?
Doesn't you want to use `$is_*` methods of dtypes?

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+      datetime_type <- dtype$`_rexpr`$get_insides()
+      out <- self$`_rexpr`$str_to_datetime(
+        format, datetime_type$tu, datetime_type$tz, strict, exact, cache, ambiguous
+      )
+      out$`_rexpr`$dt_cast_time_unit(expr, datetime_type$tu)
+    } else if (dtype == pl$Date) {
```
eitsupi, line 115: I think this comparation is not valid.
We should you `$is_date()` here.

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+    )
+  })
+}
+
+#' Get the number of bytes in strings
+#' @description
+#' Get length of the strings as UInt32 (as number of bytes). Use `$str$len_chars()`
+#' to get the number of characters.
+#'
+#' @keywords ExprStr
```
eitsupi, line 207: keywords are needed?

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+}
+
+expr_str_concat <- function(
+    delimiter = "",
+    ...,
+    ignore_nulls = TRUE) {
+  warning("$str$concat() is deprecated as of 0.18.0. Use $str$join() instead.")
+  self$`_rexpr`$str_join(delimiter, ignore_nulls) |>
+    wrap()
+}
```
eitsupi, line 268: Can we remove this?

And, please use `deprecate_warn()` instead of `warning` if you want to warn for deprecation.

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' something coercible to an Expr. Strings are parsed as column names.
+#' @details
+#' Return a copy of the string left filled with ASCII '0' digits to make a string
+#' of length width.
+#'
+#' A leading sign prefix ('+'/'-') is handled by inserting the padding after the
+#' sign character rather than before. The original string is returned if width is
+#' less than or equal to `len(s)`.
+#'
+#' @return Expr
```
eitsupi, line 404: Please inherit return from `as_polars_expr`

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' )
+expr_str_decode <- function(encoding, ..., strict = TRUE) {
+  wrap({
+    check_dots_empty0(...)
+    arg_match0(encoding, values = c("hex", "base64"))
+    if (encoding == "hex") {
+      self$`_rexpr`$str_hex_decode(strict)
+    } else if (encoding == "base64") {
+      self$`_rexpr`$str_base64_decode(strict)
+    }
```
eitsupi, line 612: Please use `switch`
https://github.com/eitsupi/neo-r-polars/blob/800cdb866057b7f45292119b2b03ba99eed474c8/R/expr-meta.R#L76-L80

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#'   pl$col("hex")$str$decode("hex")$alias("hex_decoded")$cast(pl$String)
+#' )
+expr_str_encode <- function(encoding) {
+  wrap({
+    arg_match0(encoding, values = c("hex", "base64"))
+    if (encoding == "hex") {
+      self$`_rexpr`$str_hex_encode()
+    } else if (encoding == "base64") {
+      self$`_rexpr`$str_base64_encode()
+    }
```
eitsupi, line 639: Ditto.

eitsupi (changes requested):

R/expr-string.R
etiennebacher: For now I'd rather focus on the transition from one codebase to another. This should be removed, but I'd prefer doing a last pass specifically to remove deprecated functions/args once we have transitioned.

R/expr-string.R
etiennebacher: Should this always be the case? Sometimes we specify the datatype of the Expr output, e.g. "Expr of Date, Datetime or Time type". This is useful information I think.

R/expr-string.R
etiennebacher: Not really, but that's unrelated to the switch of codebase. This can be addressed later IMO.

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' See also `$str$starts_with()` and `$str$contains()`.
+#' @inherit expr_str_contains return
+#' @examples
+#' df <- pl$DataFrame(fruits = c("apple", "mango", NA))
+#' df$select(
+#'   pl$col("fruits"),
+#'   pl$col("fruits")$str$ends_with("go")$alias("has_suffix")
+#' )
+expr_str_ends_with <- function(sub) {
+  sub <- as_polars_expr(sub, as_lit = TRUE)
```
eitsupi, line 514: Please use `wrap({})` and move this line inside of that.

R/expr-string.R
```diff
@@ -0,0 +1,1158 @@
...
+#' `List[String]` array. Contain null if original value is null or regex capture
+#' nothing.
+#'
+#' @examples
+#' df <- pl$DataFrame(foo = c("123 bla 45 asd", "xyz 678 910t"))
+#' df$select(
+#'   pl$col("foo")$str$extract_all(r"((\d+))")$alias("extracted_nrs")
+#' )
+expr_str_extract_all <- function(pattern) {
+  self$`_rexpr`$str_extract_all(as_polars_expr(pattern)$`_rexpr`)
```
eitsupi, line 689: Lack of `wrap()`

R/expr-string.R
etiennebacher: No it doesn't exist, I didn't modify those string-datetime functions too much since I'd like to address them when focusing on the date(time) expressions. However, there is no `is_datetime()` for datatype, even upstream. It shouldn't be hard to add however:
```rs
matches!(self, Datetime(_, _))
```

and actually I just realized that `is_temporal()` isn't only `Time`.

R/expr-string.R
etiennebacher: I'd rather not lose time on this since the link should be added later anyway and devtools will tell us if some links don't exist.

etiennebacher: Ah no my bad, `devtools` won't tell us

R/expr-string.R
eitsupi: Good point.
How about creat a PR to upstream polars?
(If you don't want to do, I will)

etiennebacher: I'll do it

etiennebacher: There was already some discussion in https://github.com/pola-rs/polars/issues/12196, I opened an issue here first: https://github.com/pola-rs/polars/issues/19235

R/expr-string.R
eitsupi: I understand your point, but I think there will be less friction if we stick to Python Polars clones.

R/expr-string.R
eitsupi: My expectation is that `Expr.str.concat` will be removed by Python Polars sooner than this becomes the default for r-polars, so I don't see a problem with removing it at this time.

Use `deprecate_warn()` anyway.

R/expr-string.R
etiennebacher: I've removed them as I was doing the other docs changes

R/expr-string.R
```diff
@@ -0,0 +1,1138 @@
...
+    check_dots_empty0(...)
+    self$`_rexpr`$str_join(delimiter, ignore_nulls)
+  })
+}
+
+expr_str_concat <- function(
+    delimiter = "",
+    ...,
+    ignore_nulls = TRUE) {
+  lifecycle::deprecate_warn("0.18.0", what = I("$str$concat()"), with = I("$str$join()"))
```
eitsupi, line 264: This package does not have lifecycle as a dependency. Use this instead.
https://github.com/eitsupi/neo-r-polars/blob/b90688621f7d6c4e35e34574421bc63122c58f35/R/import-standalone-lifecycle.R#L129

R/expr-string.R
```diff
@@ -0,0 +1,1138 @@
...
+#' This method is only available with the "nightly" feature.
+#' See [polars_info()] for more details.
+#' @examplesIf polars_info()$features$nightly
+#' pl$lit(c("hello there", "HI, THERE", NA))$str$to_titlecase()$to_series()
+expr_str_to_titlecase <- function() {
+  check_feature("nightly", "in $to_titlecase():")
+
+  self$`_rexpr`$str_to_titlecase(self) |>
+    wrap()
+}
```
eitsupi, line 305: This should be comment out.

eitsupi (changes requested):

src/rust/src/expr/string.rs
```diff
@@ -0,0 +1,395 @@
...
+                .clone()
+                .str()
+                .split_inclusive(by.inner.clone())
+                .into())
+        } else {
+            Ok(self.inner.clone().str().split(by.inner.clone()).into())
+        }
+    }
+
+    fn str_split_exact(&self, by: &PlRExpr, n: usize, inclusive: bool) -> Result<Self> {
```
eitsupi, line 265: I didn't know that savvy supports usize, I looked at the savvy implementation and it seems to only support conversion from i32.
So it would be better to do the conversion from NumericScalar via `<Wrap<usize>>::try_from` for now.

@yutannihilation If savvy supports usize input, do you think it would be worth adding the following conversion?
https://github.com/eitsupi/neo-r-polars/blob/b90688621f7d6c4e35e34574421bc63122c58f35/src/rust/src/conversion/mod.rs#L202-L227

R/expr-string.R
etiennebacher: Seems like they have no interest for this, issue was closed. Using `$eq(pl$Datetime())` should do the trick I think

src/rust/src/expr/string.rs
```diff
@@ -0,0 +1,396 @@
...
+        Ok(self
+            .inner
+            .clone()
+            .str()
+            .zfill(alignment.inner.clone())
+            .into())
+    }
+
+    fn str_pad_end(&self, width: NumericScalar, fillchar: &str) -> Result<Self> {
+        let width = <Wrap<usize>>::try_from(width)?.0;
```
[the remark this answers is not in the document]
eitsupi: Since there is no equivalent in R, I think it is better to define and use something like `impl TryFrom<&str> for Wrap<char>` as like this:
https://github.com/eitsupi/neo-r-polars/blob/c26e768f5b2436b0d4ffb8a6c465672ad6a9102b/src/rust/src/conversion/mod.rs#L261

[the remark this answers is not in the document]
etiennebacher: Done, I think the wording of the message might be improved if you have some suggestions. All tests pass now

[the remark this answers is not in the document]
eitsupi: Looks great!

Notifying user which argument is the problem is what is missing in the current error handling.
These will need to be improved later, but for now it seems perfect.

src/rust/src/expr/string.rs
yutannihilation: Hmm..., I might change my mind, but no plan. `f64` and `i32` are supported because these are the actual memory representation used by R, but there's no corresponding type to `usize` in R.

eitsupi:
I really want to fix more, but I don't have much time this week, so I'm merging for now. (Maybe some of the examples will fail to run)
Thanks again for the hard works!

etiennebacher:
Thanks for all the comments. I think this requires several passes anyway since sometimes it relies on other "base" expressions that are not implemented yet. I'll improve that during the week

src/rust/src/expr/string.rs
eitsupi: Thanks for your comment.
In that case it seems quite strange that usize is supported there.
Why is usize supported as an scalar input?
I didn't know it existed because it wasn't mentioned in the book

yutannihilation: Oh..., no idea. Probably I thought it would be useful, then forgot it. So, I should remove it for consistency?

> it wasn't mentioned in the book

This is a good point... An exception makes documentation a bit complex both for readers and for authors.

eitsupi: > So, I should remove it for consistency?

If it is clear why only usize is supported, I think there is no need to remove it.
Why usize is there but not u32 or u64 is odd at first glance.

yutannihilation: An integer can be intended as an index. So, usize is supported. u32 and u64 is not the case.

eitsupi: In that case, is there any reason not to allow conversion from R double to usize, as I suspect integer32 is too small to cover the full range of usize.

yutannihilation: This is just for convenience for casual usages. An ordinary R user doesn't index a vector longer than 2^32.

eitsupi: Ok, thanks.
In any case, it would be great if it were documented.

PR #8, opened 2024-10-10
feat: add more datatypes functions

etiennebacher:
This just adds `is_date()` and `is_nested_null()`, the rest is just reordering alphabetically


eitsupi:
Thank you. I think the original order was copied from py-polars, I don't think py-polars has `is_nested_null()`, do you think that is worth adding?

(I'm heading out now, so review may be 24 hours away)

etiennebacher:
Yes, I just find it easier to see what's missing by going alphabetically.

They have `is_nested_null` here: https://github.com/pola-rs/polars/blob/0870a5dbe935c98c38d35f78431e69caa3102591/crates/polars-core/src/datatypes/dtype.rs#L646

I don't have a usecase for now, it's just for completeness.

eitsupi (approved):
Thanks, let's marge for now.

eitsupi:
> They have `is_nested_null` here: [pola-rs/polars@`0870a5d`/crates/polars-core/src/datatypes/dtype.rs#L646](https://github.com/pola-rs/polars/blob/0870a5dbe935c98c38d35f78431e69caa3102591/crates/polars-core/src/datatypes/dtype.rs#L646)

It says that `is_nested_null` does not support Array of Null.
I opened pola-rs/polars#19219

PR #9, opened 2024-10-14
Add list expressions

etiennebacher:


src/rust/src/conversion/mod.rs
```diff
@@ -356,10 +357,46 @@ impl TryFrom<&str> for Wrap<WindowMapping> {
...
+        let parsed = match operation {
+            "union" => SetOperation::Union,
+            "intersection" => SetOperation::Intersection,
+            "difference" => SetOperation::Difference,
+            "symmetric_difference" => SetOperation::SymmetricDifference,
+            v => {
+                return Err(format!(
+                    "`operation` must be one of 'union', 'intersection', 'difference', 'symmetric_difference'. Got '{v}'",
+                ))
+            }
```
etiennebacher, line 381: This should be done on the R side with `arg_match()` instead, so it could maybe be `unreachable!()` here?

eitsupi: I don't think we need to use the unreachable macro, but I don't think we need to pretty the comment (perhaps something like `unexpected input` or something is good enough.)

src/rust/src/datatypes.rs
eitsupi, line 1: Are these changes needed?

eitsupi (changes requested):
Please rebase.

etiennebacher:
Aaah git is killing me

etiennebacher:
I don't have the "Merge main" or "Rebase" buttons on this repo, can you give me the rights for those?

eitsupi:
Hmmm, I don't think rebasing on GitHub is easier to do than rebasing locally, so it seems difficult to perform that operation if you are not familiar with local Git operations.
Why not abandon this branch and just copy the necessary files to the new branch?

etiennebacher:
superseded by #12

Issue #13, opened 2024-10-15
`pl$DataFrame()` should accept a `schema` argument


etiennebacher:
```r
pl$DataFrame(a = 1:3, schema = list(a = pl$Int16))
#> Error in `as_polars_df()` at neo-r-polars/R/dataframe-frame.R:55:3:
#> ! Evaluation failed.
#> Caused by error in `FUN()`:
#> ! Unsupported class for `as_polars_series()`: polars_data_type, polars_object
```

eitsupi:
To simplify the implementation, could you call the cast method later?
Since unlike Python lists, R vectors are guaranteed to have all elements of the same type and the type of the corresponding Polars is fixed, which seems more natural.

etiennebacher:
It would easier to translate code from py-polars to r-polars if we accepted this argument (it would also save some time by not having to move code to `$cast()`

eitsupi:
If we were to add an argument named `schema` to `pl$DataFrame()`, we would not be able to create a column named `schema`, so in the tidyverse world it is common to use an argument name like `.schema`.
Is that reasonable?

etiennebacher:
Yes I think it would be ok

eitsupi:
Hmmm, I looked at it a bit and it looks more complicated than I thought to implement.
I'm busy this week so I'll look at it next week.

eitsupi:
Thinking about it again, isn't there a difference between `schema` and `schema_override` here, and what you really want is `schema_override`, not `schema`, I think.
See pola-rs/r-polars#897

eitsupi:
Closed by 29ff4d2183b301985b23880e3ea3ab607533c489

PR #14, opened 2024-10-15
feat: arr namespace for expr

etiennebacher:
Needs #13 for tests


eitsupi:
> Needs #13 for tests

Please use `<dataframe>$cast()`

R/expr-array.R
```diff
@@ -0,0 +1,352 @@
...
+}
+
+#' Sum all elements in an array
+#'
+#' @inherit as_polars_expr return
+#' @examples
+#' df <- pl$DataFrame(
+#'   values = list(c(1, 2), c(3, 4), c(NA_real_, 6)),
+#'   schema = list(values = pl$Array(pl$Float64, 2))
+#' )
```
eitsupi, line 27: Could you rewrite with `cast`?

eitsupi (changes requested):
Thanks!

eitsupi (changes requested):

R/expr-array.R
```diff
@@ -0,0 +1,334 @@
...
+    wrap()
+}
+
+#' Find the maximum value in an array
+#'
+#' @inherit as_polars_expr return
+#' @inherit ExprStr_to_titlecase details
+#' @examples
+#' df <- pl$DataFrame(
+#'   values = list(c(1, 2), c(3, 4), c(NA_real_, NA_real_))
```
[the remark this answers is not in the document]
etiennebacher: That's a nice improvement

R/expr-array.R
```diff
@@ -0,0 +1,335 @@
...
+#' df$with_columns(
+#'   list = pl$col("a")$arr$to_list()
+#' )
+expr_arr_to_list <- function() {
+  self$`_rexpr`$arr_to_list() |>
+    wrap()
+}
+
+
+# TODO-REWRITE: implement this
```
eitsupi, line 309: Why not `# TODO: `?
If you use your own notation, it may not be supported by commonly used tools.
I use https://github.com/Gruntfuggly/todo-tree, but it does not support this by default.

etiennebacher: Because some TODO come from before the rewrite, so I want to be able to distinguish those from those related to the rewrite. Would it work to have `TODO: REWRITE`? I just need something that I can search for across all docs

eitsupi: Ok, thanks for clarifying.

eitsupi (changes requested):
I found that it diverged from Python Polars more than I had imagined, and I think the documentation needs to be overhauled, along with the list namespace......

I can't review any more due to lack of time.

etiennebacher:
Ok I'll do another pass on this and then on list in another PR, good catch

R/expr-array.R
```diff
@@ -17,6 +17,9 @@ namespace_expr_arr <- function(x) {
   self
 }
 
+# TODO-REWRITE: add in NEWS that some arguments in $arr$join(), $arr$sort(),
+# $arr$unique() now must be named, https://github.com/eitsupi/neo-r-polars/pull/14
```
eitsupi, line 21: It is better not to mention this in the NEWS because there are too many to mention.
If you use the positional argument, it will not malfunction because it will result in an error.

eitsupi (changes requested):
Thanks!

I think the documentation really needs an overhaul, but so far it looks like it can be merged.
(I feel that too often just reading the documentation is not what is accepted as an argument.)

etiennebacher:
> I think the documentation really needs an overhaul, but so far it looks like it can be merged.

I went through the py-polars docs and copied their docs (with some slight adaptations or a bit more info sometimes)

eitsupi:
The Python documentation assumes that type hints will be displayed, but R does not have them, so type mentions in the documentation are very important.
I feel that the explanation does not take this into account, making it difficult to guess.

R/expr-array.R
etiennebacher: Nevertheless that will convert running code to error so it should be mentioned somewhere. It could be a breaking change or a bug fix, but it should appear in the changelog IMO.

etiennebacher:
True, it would be nice to have some kind of automated stuff for that, such as adding `[IntoExpr]` first in the description of a param, that links to a page where we explain what this means. But you're right it's going to be a massive effort.

eitsupi:
Include the type in the description of all arguments. Like this:
https://github.com/eitsupi/neo-r-polars/blob/ec7b1e9c8ae609a14bc3ef0ab819c5254e034aee/R/expr-datetime.R#L31-L42

R/expr-array.R
etiennebacher: In that case would you be ok with an additional page listing all those changes so that we can focus on the most important stuff in the changelog, like https://docs.pola.rs/releases/upgrade/1 ?

I get your point, but even for us it's useful to document when a particular function changed (besides `git blame`).

etiennebacher:
I think I'm just missing your point, do you have an example of an argument whose type is unclear in this subnamespace? I agree that in other subns it can be hard to tell but here it's mostly straightforward using param description and default values.

R/expr-array.R
eitsupi: I don't think all the things like argument name changes are listed there.

Wouldn't it be better to modify the current r-polars argument names to cause breaking changes first?

R/expr-array.R
```diff
@@ -0,0 +1,418 @@
...
+# #'
+# #' df2$unnest()
+# expr_arr_to_struct <- function(fields = NULL) {
+#   self$`_rexpr`$arr_to_struct(fields) |>
+#     wrap()
+# }
+
+#' Count how often a value occurs in every sub-array
+#'
+#' @param element An expression that produces a single value.
```
eitsupi, line 353: > I think I'm just missing your point, do you have an example of an argument whose type is unclear in this subnamespace? I agree that in other subns it can be hard to tell but here it's mostly straightforward using param description and default values.

I don't think this element actually needs to be a polars expression.

R/expr-array.R
eitsupi: I think we can find out later what actually changed. Tracking everything against r-polars at this stage is a waste of time, and I think we should pay more attention to whether it is consistent with py-polars.

eitsupi: In other words, I did not look at r-polars when writing methods of each class, only py-polars, so no attention was paid to what is the same and what is different from r-polars.

R/expr-array.R
etiennebacher: Yes it is: https://docs.pola.rs/docs/python/dev/reference/expressions/api/polars.Expr.arr.count_matches.html#polars-expr-arr-count-matches

etiennebacher: Ok it could be "Can be Expr or sth coercible to an Expr"

R/expr-array.R
etiennebacher: > I think we can find out later what actually changed. 

That's gonna be very hard to find in tens of thousands of lines changed if we don't store this information beforehand. Even if not all is stored I don't think it is a bad thing to keep this info in this kind of comment. Since it's prefixed "TODO-REWRITE" it shouldn't stay hidden in the codebase anyway.

eitsupi: > That's gonna be very hard to find in tens of thousands of lines changed if we don't store this information beforehand. Even if not all is stored I don't think it is a bad thing to keep this info in this kind of comment. Since it's prefixed "TODO-REWRITE" it shouldn't stay hidden in the codebase anyway.

As for the change in argument position you pointed out, since it can be extracted with the `formals()` function, why not just write a script to do the comparison for all functions?

etiennebacher: Yes that's a good idea

R/expr-array.R
eitsupi: Thinking about it again, I don't understand what 'some TODO come from before the rewrite' means.
Can such a thing exist?

R/expr-array.R
```diff
@@ -0,0 +1,418 @@
...
+  })
+
+  class(self) <- c(
+    "polars_namespace_expr", "polars_object"
+  )
+  self
+}
+
+# TODO-REWRITE: add in NEWS that some arguments in $arr$join(), $arr$sort(),
+# $arr$unique() now must be named, https://github.com/eitsupi/neo-r-polars/pull/14
```
eitsupi, line 21:
```suggestion
```

R/expr-array.R
```diff
@@ -0,0 +1,418 @@
...
+
+# TODO-REWRITE: mention new $arr$count_matches(), $arr$explode(), $arr$first(), $arr$last(), $arr$n_unique() in NEWS
```
eitsupi, line 23:
```suggestion
```

R/expr-array.R
etiennebacher: Yes because I copy code, comments, and tests from current r-polars, and there can be existing TODO there

eitsupi: I do not understand what you mean by distinguishing between them.

- If the TODO is a problem specific to r-polars and not in py-polars, it may be removed at the stage of being copied here.
  Like this https://github.com/eitsupi/neo-r-polars/pull/19#discussion_r1805532525
- If its a TODO that is not such a problem, and common problems exist here as well, why the need to distinguish?
- How do you distinguish between the TODOs I have added here from the beginning and the ones that have been copied?

etiennebacher: It's just that in the process of rewriting I sometimes mark some stuff as `TODO-REWRITE` because we can't run them for now because they are missing some other functions. For instance we can't run `$list$concat()` because it requires `pl$concat_list()` to be implemented first. I just want to be able quickly find those things that are related to the rewriting process only. But as I said above, if this messes up with your tool for finding TODOs, I can rename those `TODO: REWRITE` or something else, I just need a string that is unique enough to search across docs.

R/expr-array.R
```diff
@@ -0,0 +1,416 @@
...
+#' @inherit as_polars_expr return
+#' @examples
+#' df <- pl$DataFrame(
+#'   a = list(c(1, 2, 3), c(4, 5, 6))
+#' )$cast(pl$Array(pl$Int64, 3))
+#' df$with_columns(last = pl$col("a")$arr$last())
+expr_arr_last <- function() {
+  self$`_rexpr`$arr_get(pl$lit(-1)$`_rexpr`, null_on_oob = TRUE) |>
+    wrap()
+}
```
eitsupi, line 403: Why not just use `self$arr$get()`?
https://github.com/pola-rs/polars/blob/be5a4b453ca74bb5e0c782c292e9fca6aa60b203/py-polars/polars/expr/array.py#L484-L507

eitsupi (changes requested):

R/expr-array.R
etiennebacher: It fails with 
```
Error in `df$with_columns()`:
! Evaluation failed in `$with_columns()`.
Caused by error:
! Evaluation failed in `$with_columns()`.
Caused by error in `pl$col("a")$arr$first()`:
! Evaluation failed in `$first()`.
Caused by error in `self$arr`:
! $ - syntax error: `arr` is not a member of
  this polars object
```

etiennebacher: Ah ok but `self$get()` works

eitsupi (approved):
Thanks. Let's merge them for now.

PR #16, opened 2024-10-16
feat: `<expr>$bin$size()` and docs of expr bin methods

etiennebacher:


R/expr-binary.R
```diff
@@ -57,3 +149,46 @@ expr_bin_encode <- function(encoding) {
...
+#' )
+expr_bin_size <- function(unit = "b") {
+  wrap({
+    arg_match0(unit, values = c(
+      "b", "bytes",
+      "kb", "kilobytes",
+      "mb", "megabytes",
+      "gb", "gigabytes",
+      "tb", "terabytes"
+    ))
```
eitsupi, line 178: It seems that only `('b', 'kb', 'mb', 'gb', 'tb')` are sufficient to support.
This is a small number and can be added to the default argument, right?
https://github.com/pola-rs/polars/blob/d8dde13cac8a2d570b2f41e16c2670b09836ee5f/py-polars/polars/expr/binary.py#L262-L263

etiennebacher: Indeed, they probably changed it recently. I'll fix that and add the choices to the default arg

eitsupi (approved):
Thanks!

PR #17, opened 2024-10-16
feat: dt namespace for expr

etiennebacher:


NAMESPACE
```diff
@@ -110,13 +110,17 @@ S3method(as_polars_series,integer64)
 S3method(as_polars_series,list)
 S3method(as_polars_series,logical)
 S3method(as_polars_series,polars_data_frame)
+S3method(as_polars_series,polars_expr)
```
eitsupi, line 113: I'm quite skeptical of this addition; Expression should be evaluated by context, and I don't feel comfortable with it suddenly being converted into a Series.

eitsupi (changes requested):

src/rust/src/functions/range.rs
```diff
@@ -0,0 +1,152 @@
...
+//         let step: i64 = step.extract()?;
+//         new_int_range::<$T>(start_v, end_v, step, PlSmallStr::from_static("literal"))
+//     });
+
+//     let s = ret.map_err(RPolarsErr::from)?;
+//     Ok(s.into())
+// }
+
+#[savvy]
+pub fn int_ranges(start: &PlRExpr, end: &PlRExpr, step: &PlRExpr, dtype: &str) -> Result<PlRExpr> {
```
eitsupi, line 47: Why is dtype `&str`?
it seems to be different from Python Polars.
https://github.com/pola-rs/polars/blob/01a4e067f50fd5f8825d4b25402336310e125d8d/crates/polars-python/src/functions/range.rs#L45-L66

eitsupi (changes requested):

NAMESPACE
etiennebacher: The problem is that without it `pl$DataFrame(x = pl$date_range(...))` fails with `Unsupported class for `as_polars_series()`: polars_expr, polars_object`

They have an `eager` argument: https://github.com/pola-rs/polars/blob/dae6cd9abd9c15667ab86e573f114c57aa43b023/py-polars/polars/functions/range/datetime_range.py#L223

but we chose to remove that to avoid having a parameter that changes the output format.

eitsupi: Is there any reason why you cannot use `pl$select()` instead of `pl$DataFrame()` there?

etiennebacher:
One issue that I have with `$str$strptime()` is that the Rust function `str_to_datetime()` should take some `Option<String>` in `format` and other args coming first, but Rust says args with `Option<>` should be last. It seems py-polars doesn't have this issue: 

https://github.com/pola-rs/polars/blob/da8e37ac9bebe8493436119d23ce25ba620b699f/crates/polars-python/src/expr/string.rs#L32

(I suppose because pyo3 reorganizes arguments).

@eitsupi do you have any idea how to fix this? The only option I have so far is to change the order of arguments (note that we will also hit this issue when implementing the I/O functions).

eitsupi:
The order of the arguments is a restriction by savvvy, and you can simply swap the order. It's already done in existing functions.
In any case, on the R side that the user touches, the order is free, so it is not a problem.

NAMESPACE
eitsupi: Are there any such use cases in existing documentation?

Adding another function something like `pl$date_range_series()` would be better, I think.

eitsupi: In my opinion, `pl$DataFrame` is usually just strict version of `pl$select`.
Wouldn't it be enough to show that to the user?

Incidentally, in the current r-polars, `pl$DataFrame` works differently than Python Polars' DataFrame constructor because it uses `pl$select` internally; Python Polars' DataFrame constructor does not accept Expr input and has no recycling rule.

eitsupi: > Perhaps this hint from Python Polars would reassure you if something like this was included in the error message?

Done by 00005bf063ec6c436cd18d5dcd018a4d1e5cf5f7

R/as_polars_series.R
```diff
@@ -463,3 +463,14 @@ as_polars_series.clock_duration <- function(x, name = NULL, ...) {
...
+# TODO-REWRITE: hack to improve
+#' @rdname as_polars_series
+#' @export
+as_polars_series.polars_expr <- function(x, name = NULL, ...) {
+  if (is.null(name)) {
+    pl$select(x)$get_columns()[[1]]$rename("")
+  } else {
+    pl$select(x)$get_columns()[[1]]$rename(name)
+  }
+}
```
eitsupi, line 476: Please remove.

eitsupi (changes requested):

NAMESPACE
eitsupi: I also added `as_polars_series.polars_expr` to ensure that such discussions are not repeated in the future.
https://github.com/eitsupi/neo-r-polars/blob/ba5e1ff3439865c8a8591ef4f8bca2e0c930f4cf/R/as_polars_series.R#L159-L169

etiennebacher: Ok the fact that `pl.DataFrame` also errors with Expr is good. 

> Adding another function something like pl$date_range_series() would be better, I think.

Not sure it's worth it if pl$select() + pl$date_range() work.

R/datatypes-classes.R
```diff
@@ -173,6 +173,16 @@ datatype__is_date <- function() {
...
 
+datatype__is_date <- function() {
+  self$`_dt`$is_date()
+}
+
+# TODO-REWRITE: do we have an equivalent of dtype == pl$Datetime, no matter the
+# unit / timezone?
+datatype__is_datetime <- function() {
+  !is.null(self$`_dt`$`_get_datatype_fields`()[["time_unit"]])
+}
```
eitsupi, line 184: I think this is wrong because Duration also has time_unit.
If we really want to support something like this, we probably need to implement something else on the Rust side.

I am imagining a method that returns sub types.

```r
dtype$`_dt`$`_dtype_name()` # c("polars_dtype_datetime", "polars_dtype_temporal")
class(dtype) # c("polars_data_type", "polars_dtype_datetime", "polars_dtype_temporal", "polars_object")
```

R/expr-string.R
```diff
@@ -107,16 +107,23 @@ expr_str_strptime <- function(
...
     check_polars_dtype(dtype)
 
-    if (pl$same_outer_dt(dtype, pl$Datetime())) {
-      datetime_type <- dtype$`_rexpr`$get_insides()
-      out <- self$`_rexpr`$str_to_datetime(
-        format, datetime_type$tu, datetime_type$tz, strict, exact, cache, ambiguous
+    if (dtype$is_datetime()) {
+      datetime_type <- dtype$`_dt`$`_get_datatype_fields`()
+      time_unit <- datetime_type[["time_unit"]]
+      time_zone <- datetime_type[["time_zone"]]
```
eitsupi, line 113: Why do you need to write it this way, isn't something like `time_unit` already bound to `dtype`?

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
+#'   as.Date("1985-01-10"),
+#'   as.difftime(2, units = "days")
+#' ) |>
+#'   as_polars_series("date")
+pl__date_range <- function(
+    start,
+    end,
+    interval = "1d",
+    ...,
+    closed = "both") {
```
eitsupi, line 42:
```suggestion
    closed = c("both", "left", "none", "right")) {
```

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
+  wrap({
+    check_dots_empty0(...)
+    arg_match0(closed, values = c("both", "left", "none", "right"))
```
eitsupi, line 45:
```suggestion
    closed <- arg_match0(closed, values = c("both", "left", "none", "right"))
```

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
+#' )
+pl__date_ranges <- function(
+    start,
+    end,
+    interval = "1d",
+    ...,
+    closed = "both") {
+  wrap({
+    check_dots_empty0(...)
+    arg_match0(closed, values = c("both", "left", "none", "right"))
```
eitsupi, line 94:
```suggestion
    closed = c("both", "left", "none", "right")) {
  wrap({
    check_dots_empty0(...)
    closed <- arg_match0(closed, values = c("both", "left", "none", "right"))
```

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
+    start,
+    end,
+    interval = "1d",
+    ...,
+    closed = "both",
+    time_unit = NULL,
+    time_zone = NULL) {
+  wrap({
+    check_dots_empty0(...)
+    arg_match0(closed, values = c("both", "left", "none", "right"))
```
eitsupi, line 151: Ditto.

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
```
eitsupi, line 199: Ditto.

R/utils-parse-duration.R
```diff
@@ -0,0 +1,106 @@
...
+#' @examples
+#' # A single character is passed as is
+#' parse_as_polars_duration_string("1d")
+#'
+#' # A single difftime is converted to a duration string
+#' parse_as_polars_duration_string(as.difftime(1, units = "days"))
+parse_as_polars_duration_string <- function(x, default = NULL, ...) {
+  if (is.null(x)) {
+    return(default)
+  }
```
eitsupi, line 48: Should be done by `parse_as_polars_duration_string.NULL`

R/utils-parse-duration.R
```diff
@@ -0,0 +1,106 @@
...
+  UseMethod("parse_as_polars_duration_string")
+}
+
+#' @exportS3Method
+parse_as_polars_duration_string.default <- function(x, default = NULL, ...) {
+  abort(
+    paste0("`", deparse(substitute(x)), "` must be a single non-NA character or difftime."),
+    call = caller_env()
+  )
```
eitsupi, line 57: Could you define a `check_` function and use it around?

eitsupi (changes requested):

R/datatypes-classes.R
eitsupi: > I am imagining a method that returns sub types.

Done in 2f3357d5887653c09b7aed10c544328961ac27e4

R/utils-parse-duration.R
etiennebacher: There are only two occurrences for this since the `.default` method errors all the time so it's not a check, and adding a check function for the remaining two makes it harder to properly replace `x` by its original expression in the error message.

eitsupi: Ok, thanks.

src/rust/src/expr/datetime.rs
```diff
@@ -29,4 +29,143 @@ impl PlRExpr {
...
+    fn dt_combine(&self, time: &PlRExpr, time_unit: &str) -> Result<Self> {
+        Ok(self
+            .inner
+            .clone()
+            .dt()
+            .combine(time.inner.clone(), <Wrap<TimeUnit>>::try_from(time_unit)?.0)
+            .into())
+    }
+
+    fn dt_strftime(&self, fmt: &str) -> Result<Self> {
```
eitsupi, line 54: Argument name mismatch.

src/rust/src/expr/datetime.rs
```diff
@@ -29,4 +29,143 @@ impl PlRExpr {
...
+        Ok(self.clone().inner.dt().millisecond().into())
+    }
+    fn dt_microsecond(&self) -> Result<Self> {
+        Ok(self.clone().inner.dt().microsecond().into())
+    }
+    fn dt_nanosecond(&self) -> Result<Self> {
+        Ok(self.clone().inner.dt().nanosecond().into())
+    }
+
+    fn dt_timestamp(&self, tu: &str) -> Result<Self> {
```
eitsupi, line 100: Argument name mismatch.

src/rust/src/expr/datetime.rs
```diff
@@ -29,4 +29,143 @@ impl PlRExpr {
...
+                    s.take_materialized_series()
+                        .timestamp(TimeUnit::Milliseconds)
+                        .map(|ca| Some((ca / 1000).into_column()))
+                },
+                GetOutput::from_type(DataType::Int64),
+            )
+            .into())
+    }
+
+    fn dt_with_time_unit(&self, tu: &str) -> Result<Self> {
```
eitsupi, line 124: Argument name mismatch.

src/rust/src/expr/datetime.rs
```diff
@@ -29,4 +29,143 @@ impl PlRExpr {
...
+        Ok(self
+            .inner
+            .clone()
+            .dt()
+            .with_time_unit(<Wrap<TimeUnit>>::try_from(tu)?.0)
+            .into())
+    }
+
+    fn dt_cast_time_unit(&self, tu: &str) -> Result<Self> {
```
eitsupi, line 133: Argument name mismatch.

eitsupi (changes requested):

src/rust/src/expr/datetime.rs
```diff
@@ -29,4 +29,188 @@ impl PlRExpr {
...
+        &self,
+        n: &PlRExpr,
+        week_mask: LogicalSexp,
+        holidays: IntegerSexp,
+        roll: &str,
+    ) -> Result<Self> {
+        let week_mask: [bool; 7] = week_mask
+            .to_vec()
+            .try_into()
+            .unwrap_or_else(|_| !unreachable!());
```
etiennebacher, line 206: Maybe there's a better way for this, but it seemed simpler to check that `week_mask` is made of 7 booleans on the R side and then put `!unreachable` here

R/expr-datetime.R
```diff
@@ -82,3 +114,971 @@ expr_dt_replace_time_zone <- function(
...
+    roll = c("raise", "backward", "forward")) {
+  wrap({
+    roll <- arg_match0(roll, values = c("raise", "backward", "forward"))
+    unix_epoch <- as.Date("1970-1-1")
+    if (length(n) == 1 && is.numeric(n) && as.integer(n) == n) {
+      n <- as.integer(n)
+    }
+    if (!is.logical(week_mask) || anyNA(week_mask) || length(week_mask) != 7) {
+      abort("`week_mask` must be a vector with 7 logical values, without any NA.")
+    }
```
eitsupi, line 1076: How about `is_logical(week_mask, n = 7L)`?

src/rust/src/expr/datetime.rs
eitsupi: I think that makes sense.
But can't we just use something like `.map_err(savvy::Error::from)` here to avoid panic?

eitsupi (changes requested):

src/rust/src/expr/datetime.rs
etiennebacher: If I use it instead of the unwrap_or_else call then I have "the trait bound `savvy::Error: From<Vec<bool>>` is not satisfied"

R/functions-range.R
```diff
@@ -0,0 +1,207 @@
...
+#' pl$date_range(as.Date("2022-01-01"), as.Date("2022-03-01"), "1mo") |>
+#'   as_polars_series("date")
+#'
+#' # Using `difftime` object to specify the interval:
+#' pl$date_range(
+#'   as.Date("1985-01-01"),
+#'   as.Date("1985-01-10"),
+#'   as.difftime(2, units = "days")
+#' ) |>
+#'   as_polars_series("date")
```
eitsupi, line 36: Could you update examples?

R/expr-string.R
```diff
@@ -106,16 +106,21 @@ expr_str_strptime <- function(
   wrap({
     check_dots_empty0(...)
     check_polars_dtype(dtype)
-
     dtype_class <- class(dtype)
     if ("polars_dtype_datetime" %in% dtype_class) {
+      # TODO-REWRITE: why is this needed?
+      if (dtype$time_unit == "μs") {
+        dtype$time_unit <- "us"
+      }
```
eitsupi, line 114:
```suggestion
```

R/expr-string.R
```diff
@@ -194,7 +201,9 @@ expr_str_to_datetime <- function(
   wrap({
     check_dots_empty0(...)
     self$`_rexpr`$str_to_datetime(
-      self, format, time_unit, time_zone, strict, exact, cache, ambiguous
+      format = format, time_unit = time_unit, time_zone = time_zone,
+      strict = strict, exact = exact, cache = cache,
+      ambiguous = as_polars_expr(ambiguous, as_lit = TRUE)$`_rexpr`
```
eitsupi, line 206: Please check ambiguous as like this:
https://github.com/eitsupi/neo-r-polars/blob/49ac30a9bf6f6d88b3b0725d08b59b75c0244c80/R/series-to_r_vector.R#L191-L194

eitsupi (changes requested):

R/expr-string.R
etiennebacher: Something like this is required, otherwise this fails:

```r
pl$DataFrame(x = "2020-01-01 01:00Z")$select(pl$col("x")$str$strptime(pl$Datetime()))

Error:
Caused by error:
Caused by error in `pl$col("x")$str$strptime()`:
! Evaluation failed in `$strptime()`.
Caused by error:
! unsupported value: 'μs'
```
If you prefer I can handle this in `impl TryFrom<&str> for Wrap<TimeUnit>` to have both "us" and "μs" return `TimeUnit::Microseconds`.

eitsupi: Ah, I did not know that it returns such value!
This is a bug and will be fixed soon.

R/expr-datetime.R
```diff
@@ -82,3 +114,973 @@ expr_dt_replace_time_zone <- function(
...
+#'     as.POSIXct("2001-01-01 00:00"),
+#'     as.POSIXct("2001-01-01 01:00"),
+#'     as.difftime("0:10:0")
+#'   )
+#' )
+#' df$with_columns(round = pl$col("datetime")$dt$round("1h"))
+expr_dt_round <- function(every) {
+  every <- parse_as_polars_duration_string(every, default = "0ns")
+  self$`_rexpr`$dt_round(as_polars_expr(every, as_lit = TRUE)$`_rexpr`) |>
+    wrap()
```
eitsupi, line 205: Please `wrap()` them.

R/expr-datetime.R
```diff
@@ -82,3 +114,973 @@ expr_dt_replace_time_zone <- function(
...
+#' )
+#'
+#' df$with_columns(
+#'   second = pl$col("datetime")$dt$second(),
+#'   second_fractional = pl$col("datetime")$dt$second(fractional = TRUE)
+#' )
+expr_dt_second <- function(fractional = FALSE) {
+  wrap({
+    sec <- self$`_rexpr`$dt_second()
+    if (fractional) {
```
eitsupi, line 523:
```suggestion
    if (isTRUE(fractional)) {
```

R/expr-datetime.R
```diff
@@ -82,3 +114,973 @@ expr_dt_replace_time_zone <- function(
...
+expr_dt_epoch <- function(time_unit = c("us", "ns", "ms", "s", "d")) {
+  wrap({
+    time_unit <- arg_match0(time_unit, values = c("us", "ns", "ms", "s", "d"))
+    switch(time_unit,
+      "ms" = ,
+      "us" = ,
+      "ns" = self$`_rexpr`$dt_timestamp(time_unit),
+      "s" = self$`_rexpr`$dt_epoch_seconds(),
+      "d" = self$`_rexpr`$cast(pl$Date$`_dt`, strict = TRUE, wrap_numerical = FALSE)$cast(pl$Int32$`_dt`, strict = TRUE, wrap_numerical = FALSE)
+    )
```
eitsupi, line 625: Could you add `abort()` at the last of switch?

R/expr-string.R
```diff
@@ -106,16 +106,17 @@ expr_str_strptime <- function(
...
     check_polars_dtype(dtype)
-
     dtype_class <- class(dtype)
     if ("polars_dtype_datetime" %in% dtype_class) {
       self$`_rexpr`$str_to_datetime(
-        format, dtype$time_unit, dtype$time_zone, strict, exact, cache, ambiguous
-      )$dt_cast_time_unit(expr, dtype$time_unit)
+        format = format, time_unit = dtype$time_unit, time_zone = dtype$time_zone,
+        strict = strict, exact = exact, cache = cache,
+        ambiguous = as_polars_expr(ambiguous, as_lit = TRUE)$`_rexpr`
```
eitsupi, line 114: Update as like `expr_str_to_datetime`

R/expr-string.R
```diff
@@ -193,8 +196,14 @@ expr_str_to_datetime <- function(
...
     check_dots_empty0(...)
+    if (!is_polars_expr(ambiguous)) {
+      ambiguous <- arg_match0(ambiguous, c("raise", "earliest", "latest", "null")) |>
+        as_polars_expr(as_lit = TRUE)
+    }
     self$`_rexpr`$str_to_datetime(
-      self, format, time_unit, time_zone, strict, exact, cache, ambiguous
+      format = format, time_unit = time_unit, time_zone = time_zone,
+      strict = strict, exact = exact, cache = cache,
+      ambiguous = as_polars_expr(ambiguous, as_lit = TRUE)$`_rexpr`
```
eitsupi, line 206:
```suggestion
      ambiguous = ambiguous$`_rexpr`
```

R/functions-range.R
```diff
@@ -0,0 +1,213 @@
...
+#' pl$select(
+#'   date = pl$date_range(
+#'     as.Date("1985-01-01"),
+#'     as.Date("1985-01-10"),
+#'     as.difftime(2, units = "days")
+#'   )
+#' )
+pl__date_range <- function(
+    start,
+    end,
```
eitsupi, line 41: Could you check types of these objects?
Perhaps a dedicated function would be worthwhile.

eitsupi (changes requested):

tests/testthat/test-functions-range.R
etiennebacher, line 1: I only added a few snapshots for them since they are used extensively in tests for expr-date

eitsupi (approved):
Thanks for the huge work!

I'd like to merge this before I go to bed, can I merge it right away?

PR #18, opened 2024-10-16
docs: add some docs for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -10,7 +10,7 @@
 #' This function is basically a shortcut for `list(...) |> as_polars_df()`,
 #' so each argument in `...` is converted to a Polars Series by [as_polars_series()]
 #' and then passed to [as_polars_df()].
-#' @aliases polars_data_frame DataFrame
```
eitsupi, line 13: Why do you remove this?

R/dataframe-frame.R
```diff
@@ -112,16 +112,55 @@ dataframe__to_struct <- function(name = "") {
...
+#' pl$DataFrame(iris)$lazy()
 dataframe__lazy <- function() {
   self$`_df`$lazy() |>
     wrap()
 }
 
+#' Clone a DataFrame
+#'
+#' This makes a very cheap deep copy/clone of an existing
+#' [`DataFrame`][DataFrame_class]. Rarely useful as `DataFrame`s are nearly 100%
```
eitsupi, line 129: `DataFrame_class` does not exist.
`[DataFrame]` should be enough here.

eitsupi (changes requested):

R/dataframe-frame.R
etiennebacher: I don't think this is useful and it's not used consistently across docs. I can add it back if you prefer having it, but I just never use aliases (I don't even know how users can see the list of aliases)

etiennebacher: Ok, so that is limited to `pl__DataFrame`, `pl__LazyFrame`, `pl__Series` and the expr page? Or you wanted to add aliases to other functions as well?

eitsupi: Yes, my intention was to be able to use help with typical class names (series, expr, dataframe, ...).
So I don't think alias is necessary for functions/methods.

etiennebacher: Ok I'll revert this change

R/dataframe-frame.R
```diff
@@ -112,16 +113,55 @@ dataframe__to_struct <- function(name = "") {
     wrap()
 }
 
+#' Convert an existing DataFrame to a LazyFrame
+#' @description Start a new lazy query from a DataFrame.
+#'
+#' @inherit as_polars_lf return
+#' @examples
+#' pl$DataFrame(iris)$lazy()
```
eitsupi, line 121: The behavior of `pl$DataFrame(<data.frame>)` has been changed.
Could you copy Python example?
https://github.com/pola-rs/polars/blob/577cd62d30361a0e03a018e8761aae7dceedd784/py-polars/polars/dataframe/frame.py#L8884-L8919

R/dataframe-frame.R
```diff
@@ -112,16 +113,55 @@ dataframe__to_struct <- function(name = "") {
...
 dataframe__lazy <- function() {
   self$`_df`$lazy() |>
     wrap()
 }
 
+#' Clone a DataFrame
+#'
+#' This makes a very cheap deep copy/clone of an existing
+#' [`DataFrame`]. Rarely useful as `DataFrame`s are nearly 100%
```
eitsupi, line 130:
```suggestion
#' [DataFrame]. Rarely useful as [DataFrame]s are nearly 100%
```

R/dataframe-frame.R
```diff
@@ -112,16 +113,55 @@ dataframe__to_struct <- function(name = "") {
...
+#' @examples
+#' pl$DataFrame(iris)$lazy()
 dataframe__lazy <- function() {
   self$`_df`$lazy() |>
     wrap()
 }
 
+#' Clone a DataFrame
+#'
+#' This makes a very cheap deep copy/clone of an existing
```
eitsupi, line 129: Could you copy from Python Polars?
https://github.com/pola-rs/polars/blob/577cd62d30361a0e03a018e8761aae7dceedd784/py-polars/polars/dataframe/frame.py#L7762-L7795

R/dataframe-frame.R
```diff
@@ -112,16 +113,55 @@ dataframe__to_struct <- function(name = "") {
...
+#' give_attr <- function(data) {
+#'   data <- data$clone()
+#'   attr(data, "created_on") <- "2024-01-29"
+#'   data
+#' }
+#' df1 <- pl$DataFrame(iris)
+#' df2 <- give_attr(df1)
+#'
+#' # now, the original DataFrame doesn't get this attribute
+#' attributes(df1)
```
eitsupi, line 158: I think these examples are not good...

R/dataframe-frame.R
```diff
@@ -112,16 +113,55 @@ dataframe__to_struct <- function(name = "") {
...
 dataframe__clone <- function() {
   self$`_df`$clone() |>
     wrap()
 }
 
+# TODO-REWRITE: fix link to Series
```
eitsupi, line 164: Is the link not working properly?

R/dataframe-frame.R
```diff
@@ -145,25 +185,132 @@ dataframe__get_columns <- function() {
...
+#' @param ... Columns to keep. Those can be expressions (e.g `pl$col("a")`),
+#' column names  (e.g `"a"`), or list containing expressions or column names
+#' (e.g `list(pl$col("a"))`).
+#'
+#' @inherit as_polars_df return
+#' @examples
+#' pl$DataFrame(iris)$select(
+#'   pl$col("Sepal.Length")$abs()$alias("abs_SL"),
+#'   (pl$col("Sepal.Length") + 2)$alias("add_2_SL")
+#' )
```
eitsupi, line 238: I think these methods need to be inherited from lazyframe's.

R/dataframe-frame.R
etiennebacher: This is actually the only case that I could find where `clone()` is useful (the python docs don't explain at all why one would need to use it).

eitsupi: The behavior of renaming a column using `.columns = ` in Python Polars is a good example, but that is not implemented here (instead we have this)...

https://github.com/eitsupi/neo-r-polars/blob/557325983d62686c0d35dc820e0ff2af1310e916/R/dataframe-frame.R#L88-L95

Unlike the current R polars, here the DataFrame is an environment object, so how about an example like this? (i.e., environments are one of the few objects in R that are passed by reference)

```r
df_a <- pl$DataFrame(x = 1)
df_b <- df_a
df_c <- df_a$clone()
rlang::env_label(df_a)
rlang::env_label(df_b)
rlang::env_label(df_c)
```

R/dataframe-frame.R
```diff
@@ -112,19 +113,57 @@ dataframe__to_struct <- function(name = "") {
...
 dataframe__clone <- function() {
   self$`_df`$clone() |>
     wrap()
 }
 
-#' Get the DataFrame as a List of Series
+#' Get the DataFrame as a list of Series
 #'
-#' @return A [list] of [Series]
+#' @return A list of [Series]
```
eitsupi, line 166: Why you edit this?

eitsupi (changes requested):

R/dataframe-frame.R
etiennebacher: I don't see the point of linking base R functions. We don't list `vector`, `logical`, etc.

R/dataframe-frame.R
etiennebacher: Yes but that doesn't show why one would need this. I don't mind putting this `rlang::env_label()` before the other example to show that the address changes, but I'd also like to show a valid usecase for this and I think the one with attributes does the job. What don't you like about it? I don't mind replacing them if you have some alternative usecases

R/dataframe-frame.R
eitsupi: The term `list` is confusing with Polars list dtype, etc.

R/dataframe-frame.R
eitsupi: What is the reason for adding the attribute?
I don't understand the use case for adding attributes without defining another class.
If you add an attribute, it will disappear immediately upon some operation.

R/dataframe-frame.R
etiennebacher: In the docs we use "A polars ...", e.g. "A polars expression". I think it's clear enough that if we don't then we refer to standard R terms.

eitsupi: Ok, that makes sense.

R/dataframe-frame.R
etiennebacher: Here's how I use it in `group_by`: https://github.com/etiennebacher/tidypolars/blob/main/R/groups.R#L31

eitsupi: Anyway, since this is the behavior of environment types in general, not just polars object, can you add examples of `rlang::env_label` first?
Then as a developmental example you can explain why it matters.

etiennebacher: > Will using attributes affect print results, etc.?

I have a custom print method for `tidypolars` objects so I handle attributes there.

I'll add the rlang::env_label() examples

R/dataframe-frame.R
etiennebacher: I added inherited docs for select and with_columns, but other methods in lazyframe don't have their docs yet so I'd like to add those and deduplicate the dataframe/lazyframe in a future PR

R/lazyframe-frame.R
```diff
@@ -264,14 +267,14 @@ lazyframe__sort <- function(
   })
 }
 
-#' Add columns to this LazyFrame
+#' Modify/append column(s) of a LazyFrame
 #'
-#' Added columns will replace existing columns with the same name.
+#' Add columns or modify existing ones with expressions. This is
+#' the equivalent of `dplyr::mutate()` as it keeps unmentioned columns (unlike
+#' `$select()`).
```
eitsupi, line 274: One major difference from `dplyr::mutate` is that in `mutate` you can reference added columns within the same function, which is not possible here.

R/lazyframe-frame.R
```diff
@@ -264,14 +267,14 @@ lazyframe__sort <- function(
...
 #'
-#' Creating a new LazyFrame using this method does not create a new copy of
-#' existing data.
-#' @inherit pl__LazyFrame return
 #' @inheritParams lazyframe__select
+#' @inherit pl__LazyFrame return
```
eitsupi, line 277: Why this change?
I think that inherit should be above inheritParams because it can inherit including params.

R/dataframe-frame.R
```diff
@@ -201,22 +379,86 @@ dataframe__tail <- function(n = 5) {
...
+#' Cast DataFrame column(s) to the specified dtype
+#'
+#' @inherit LazyFrame_cast description params
+#'
+#' @inherit as_polars_df return
+#' @examples
+#' df <- pl$DataFrame(
+#'   foo = 1:3,
+#'   bar = c(6, 7, 8),
+#'   ham = as.Date(c("2020-01-02", "2020-03-04", "2020-05-06"))
```
eitsupi, line 409:
```suggestion
#'   ham = as.Date(c("2020-01-02", "2020-03-04", "2020-05-06")),
```

R/dataframe-frame.R
```diff
@@ -201,22 +379,86 @@ dataframe__tail <- function(n = 5) {
   })
 }
 
+#' Drop columns of a DataFrame
+#'
+#' @param ... Characters of column names to drop. Passed to [`pl$col()`][pl_col].
```
eitsupi, line 384: Please use dynamic-dots.
See select's document.

eitsupi (changes requested):

R/dataframe-frame.R
etiennebacher: Why? I really don't want to change this by hand across all docs

R/lazyframe-frame.R
etiennebacher: I put the inherit first but I really don't think that matters since we didn't specify params in the inherit

R/dataframe-frame.R
eitsupi: For consistency with the Python examples, but certainly not of great importance. Sorry.

R/dataframe-frame.R
```diff
@@ -127,19 +128,67 @@ dataframe__to_struct <- function(name = "") {
...
+#'
+#' This is a cheap operation that does not copy data. This can be useful when
+#' dealing with attributes (see examples).
+#'
+#' @inherit as_polars_df return
+#' @examples
+#' df1 <- as_polars_df(iris)
+#'
+#' # Assigning doesn't change the memory address of the environment, but
+#' # cloning does.
```
eitsupi, line 152: I think it is important to note that the environment object containing the DataFrame is not copied by assignment because it has *reference semantics*.
https://adv-r.hadley.nz/environments.html#explicit-envs

etiennebacher: Does "Assigning doesn't change the memory address of the environment because it has reference semantics, but cloning does." work for you?

If not can you suggest a replacement?

eitsupi: How about that?
> Assigning does not copy the DataFrame (environment object). This is because environment objects have reference semantics.
> Calling `$clone()` creates a new environment.

etiennebacher: Sounds good, I think it fits better in the description than in the examples though. I'll move this

eitsupi (approved):
Thanks. Let's merge and move on.

PR #19, opened 2024-10-16
feat: add some docs and missing functions for Expr

etiennebacher:


NAMESPACE
```diff
@@ -35,7 +35,7 @@ S3method("-",polars_expr)
 S3method("-",polars_series)
 S3method("/",polars_expr)
 S3method("/",polars_series)
-S3method("<",polras_expr)
+S3method("<",polars_expr)
```
eitsupi, line 38: Good catch.

R/expr-expr.R
```diff
@@ -70,195 +69,523 @@ pl__deserialize_expr <- function(data, ..., format = c("binary", "json")) {
...
+#' df <- pl$DataFrame(
+#'   x = -2:2,
+#'   y = c(0.5, 0, 0, -4, -0.5)
+#' )
+#'
+#' df$with_columns(
+#'   `x/2` = pl$col("x")$div(2),
+#'   `x/y` = pl$col("x")$div(pl$col("y"))
+#' )
+expr__truediv <- function(other) {
```
eitsupi, line 165: Why did you change the name to `truediv`?
IIUC this has a different name in Rust and Python, so we have to choose which one we prefer.
I think with underscore makes more sense here.

eitsupi: `truediv` comes from Python. Rust and R do not have to choose this.
https://docs.python.org/3/library/operator.html

etiennebacher: I think we should just match py-polars and rename `floor_div` to `floordiv`, and use `truediv`. If we had `floor_div()` / `true_div()` in R, we probably would have matched those but given that those don't exist I think we should prioritize consistency between r-polars and py-polars functions. It would just be annoying to have an error when converting R <-> python code just because of an underscore.

> Python Polars cannot use not(), so it adopts not_(), and so on, which is a limitation of the Python world of naming.

IMO that's a different issue. For division, they could have used `floor_div()` and `true_div()` but just decided to stick to the most familiar in python (unless I missed something).

eitsupi: Ok. I think it is not a bad idea to follow Python here.

R/expr-expr.R
```diff
@@ -6,12 +6,11 @@
 #' An expression is a tree of operations that describe how to construct one or more [Series].
 #' As the outputs are [Series], it is straightforward to apply a sequence of expressions each of
 #' which transforms the output from the previous step.
-#' See examples for details.
-#' @name polars_expr
-#' @aliases Expr expression
```
eitsupi, line 11: Please do not remove the aliases.
Same as https://github.com/eitsupi/neo-r-polars/pull/18#discussion_r1804634224

R/expr-expr.R
```diff
@@ -70,195 +69,532 @@ pl__deserialize_expr <- function(data, ..., format = c("binary", "json")) {
   })
 }
 
+#' Add two expressions
+#'
+#' Method equivalent of addition operator `expr + other`.
+#' @param other numeric or string value; accepts expression input.
+#' @return [Expr][expr__class]
+#' @seealso
```
eitsupi, line 77: Please inherits return from as_polars_expr

R/expr-expr.R
```diff
@@ -70,195 +69,532 @@ pl__deserialize_expr <- function(data, ..., format = c("binary", "json")) {
...
   wrap({
     other <- as_polars_expr(other, as_lit = TRUE)
     self$`_rexpr`$add(other$`_rexpr`)
   })
 }
 
+#' Substract two expressions
+#'
+#' Method equivalent of subtraction operator `expr - other`.
+#' @inherit expr__truediv params return
```
eitsupi, line 106: Ditto.

R/expr-expr.R
```diff
@@ -364,12 +893,109 @@ expr__over <- function(
...
+#' )
 expr__filter <- function(...) {
   parse_predicates_constraints_into_expression(...) |>
     self$`_rexpr`$filter() |>
     wrap()
 }
 
+## TODO Better explain aggregate list
+
+#' Map an expression with an R function
```
eitsupi, line 925: Set noRd for now as map_batches is broken and will probabilistically stop the workflow when run on CI.

eitsupi (changes requested):
I don't have time to review everything, but a few comments.

R/expr-expr.R
```diff
@@ -268,20 +618,79 @@ expr__cast <- function(dtype, ..., strict = TRUE, wrap_numerical = FALSE) {
...
+#'   a = c(6, 1, 0, NA, Inf, NaN)
+#' )$with_columns(arg_sorted = pl$col("a")$arg_sort())
 expr__arg_sort <- function(..., descending = FALSE, nulls_last = FALSE) {
   wrap({
     check_dots_empty0(...)
     self$`_rexpr`$arg_sort(descending, nulls_last)
   })
 }
 
+# TODO: rewrite `by` to `...` <https://github.com/pola-rs/r-polars/pull/997>
```
eitsupi, line 657:
```suggestion
```

R/expr-expr.R
```diff
@@ -268,20 +618,79 @@ expr__cast <- function(dtype, ..., strict = TRUE, wrap_numerical = FALSE) {
   })
 }
 
+#' Sort an Expr
+#'
+#' Sort this column. If used in a groupby context, the groups are sorted.
+#'
+#' @inheritParams Series_sort
```
eitsupi, line 625: Please inherits dots empty

eitsupi (changes requested):
Sorry, no time for a review. As a whole, could you please specify the type?
For example, there are arguments that are ambiguous as to whether only numbers are accepted or whether expr is also acceptable.
Please refer to the existing documentation I have written.

eitsupi:
I have noticed that the current conversion from Date class to Series is incorrect. (See f2005f38b896a24bb53f568fa390f27be728a994)
I need the `floor` method to fix this, would you add it in this PR?

etiennebacher:
Nice catch!

>  I need the floor method to fix this, would you add it in this PR?

Yes I will. The workflow I have now is that I review all exprs from py-polars API page so I'll include all missing expressions and review docs for all of them in this PR. Is this ok for you? This might take a few days to get done, so you might want to add the floor method yourself to fix the other issue

eitsupi:
> Yes I will. The workflow I have now is that I review all exprs from py-polars API page so I'll include all missing expressions and review docs for all of them in this PR. Is this ok for you? This might take a few days to get done, so you might want to add the floor method yourself to fix the other issue

Excellent, thank you. I'm not in much of a hurry so will probably fix it after this PR is merged.

eitsupi:
> I need the `floor` method to fix this

I thought about it again, and since extra processing such as converting to a DataFrame occurs when using floor, I wrote a new function on the Rust side and fixed it (2165f6f3edbaae6e0e7aed667af5e001161edd09)

R/expr-expr.R
```diff
@@ -70,195 +70,591 @@ pl__deserialize_expr <- function(data, ..., format = c("binary", "json")) {
...
   self$`_rexpr`$not() |>
     wrap()
 }
 
 # Beacuse the $not method and the $invert method are distinguished in the selector,
 # this is only necessary to map the $invert method to the `!` operator.
-expr__invert <- expr__not
 
+expr__invert <-
+  expr__not
```
eitsupi, line 445: Why is this change needed?

R/expr-expr.R
```diff
@@ -390,49 +1102,2934 @@ expr__map_batches <- function(
...
+#' @inherit as_polars_expr return
+#' @details
+#' The Dtypes Int8, UInt8, Int16 and UInt16 are cast to Int64 before summing to
+#' prevent overflow issues.
+#' @examples
+#' pl$DataFrame(a = 1:4)$with_columns(
+#'   cum_sum = pl$col("a")$cum_sum(),
+#'   cum_sum_reversed = pl$col("a")$cum_sum(reverse = TRUE)
+#' )
+expr__cum_sum <- function(reverse = FALSE) {
```
eitsupi, line 1288: Please add `...`

R/expr-expr.R
```diff
@@ -390,49 +1102,2934 @@ expr__map_batches <- function(
...
+#'
+#' @param maintain_order Maintain order of data. This requires more work.
+#'
+#' @inherit as_polars_expr return
+#' @examples
+#' df <- pl$DataFrame(a = c(1, 1, 2))
+#' df$select(pl$col("a")$unique())
+expr__unique <- function(..., maintain_order = FALSE) {
+  wrap({
+    rlang::check_dots_empty0(...)
```
eitsupi, line 2277:
```suggestion
    check_dots_empty0(...)
```

R/expr-expr.R
```diff
@@ -390,49 +1102,2934 @@ expr__map_batches <- function(
...
+#'   )
+#' )
+#'
+#' # Center the values in the window
+#' df$with_columns(
+#'   rolling_quantile = pl$col("a")$rolling_quantile(
+#'     quantile = 0.25, window_size = 5, center = TRUE
+#'   )
+#' )
+expr__rolling_quantile <- function(
```
eitsupi, line 2576: Please add `experimental` badge.

eitsupi (changes requested):

R/expr-expr.R
eitsupi: On second thought, I still think this method name should be changed to `true_div` and `floor_div` as it seems odd even for Python Polars (pola-rs/polars#19531).

R/as_polars_expr.R
```diff
@@ -207,7 +207,7 @@ as_polars_expr.integer <- function(x, ...) {
 as_polars_expr.double <- function(x, ...) {
   wrap({
     if (length(x) == 1L) {
-      if (is.na(x)) {
+      if (is.na(x) && !is.nan(x)) {
```
eitsupi, line 210: Good catch, I will fix this in the main.

eitsupi: Fixed by 728a97e9942dca551dc3a18bf3b9089a81678ebf

tests/testthat/test-expr-expr.R
```diff
@@ -29,3 +29,2946 @@ test_that("map_batches works", {
...
+
+  # clip() works with temporal
+  df <- pl$DataFrame(foo = as.Date(c("2020-01-01", "2020-01-02")))
+  expect_equal(
+    df$select(clipped = pl$col("foo")$clip(lower_bound = pl$lit("2020-01-02"))),
+    pl$DataFrame(clipped = as.Date(c("2020-01-02", "2020-01-02")))
+  )
+})
+
+# TODO-REWRITE: lower bound i32 becomes NA once converted to R
```
etiennebacher, line 2085: Don't know how to fix this

tests/testthat/test-expr-expr.R
```diff
@@ -29,3 +29,2946 @@ test_that("map_batches works", {
...
+    pl$DataFrame(
+      kurt_TT = R_kurtosis(l2$a, TRUE, TRUE),
+      kurt_TF = R_kurtosis(l2$a, TRUE, FALSE),
+      kurt_FT = R_kurtosis(l2$a, FALSE, TRUE),
+      kurt_FF = R_kurtosis(l2$a, FALSE, FALSE)
+    )
+  )
+})
+
+# TODO-REWRITE: panicks when one of the bounds is NaN
```
etiennebacher, line 1979: Don't know why this occurs, it throws a normal error in python

R/expr-expr.R
eitsupi: @etiennebacher After all, this seems to be Python-only, so wouldn't it make more sense to prefer a naming convention that uses underscores in R?
https://github.com/pola-rs/polars/issues/19531#issuecomment-2449458291

etiennebacher: I still think the same as above: given that `floordiv`/`truediv` don't exist in R, I don't see a reason for diverging from the py-polars syntax (especially since the syntax of function names in R varies a lot).

R/utils-various.R
```diff
@@ -16,3 +16,32 @@ extend_bool <- function(value, n_match, value_name, match_name) {
     value
   }
 }
+
+#' Used in ewm_* functions
+#' @noRd
+prepare_alpha <- function(com, span, half_life, alpha) {
+  check_exclusive(com, span, half_life, alpha, .call = caller_env())
```
eitsupi, line 23: Don't we need to check the type of each argument?

R/expr-expr.R
eitsupi: So what I am saying is that since this is not Python, shouldn't Polars naming convention take precedence over Python? See this comment.
https://github.com/pola-rs/polars/issues/6120#issuecomment-1374911565

In fact, I don't think Rust follows Python's naming conventions.

eitsupi (changes requested):

tests/testthat/test-expr-expr.R
eitsupi: How about cast to i64?

tests/testthat/test-expr-expr.R
eitsupi: Sorry, but this test is quite hard to read. I don't know what you are trying to do. Could you please rewrite it?

eitsupi (changes requested):

R/expr-expr.R
etiennebacher: I get your point, but I think it's more important to have consistency with the actual functions implemented in py-polars. It would be weird to say that we have this inconsistency because we follow the polars naming convention while the original doesn't.

tests/testthat/test-expr-expr.R
etiennebacher: Do you mean in the test or in the implementation?

If we do this in the tests only, the issue is that users might silently run into this int32 -> NA conversion. 
If we do this in the implementation this would add confusion since the input would be int32 but the output would be int64.

Should we have a warning when converting int32 lower bound to R since this is the only case this int32 -> NA conversion would happen?

eitsupi: I understand the need for this, but do not plan to address it immediately. So I think we need to cast to String or I64 during testing.
https://github.com/eitsupi/neo-r-polars/blob/1d7ffa8b44966baffc14524d1498938a0cf119f9/src/rust/src/series/export.rs#L68

R/expr-expr.R
eitsupi: In that case, why not use true_div and floor_div as canonical (i.e., we need to use this one in the document) and also add truediv and floordiv as aliases for compatibility with Python Polars?

tests/testthat/test-expr-expr.R
```diff
@@ -1976,104 +1976,79 @@ test_that("kurtosis", {
...
+  # expect_equal(
+  #   df$select(
+  #     clip_min_int = pl$col("int")$clip(NaN),
+  #     clip_min_float = pl$col("float")$clip(NaN)
+  #   ),
+  #   pl$DataFrame(
+  #     clip_min_int = c(1:3, NA, -.Machine$integer.max, .Machine$integer.max),
+  #     clip_min_float = c(1, 2, 3, NA, -Inf, Inf)
+  #   )
+  # )
```
eitsupi, line 2050: It seems that these are only panic for dev profiles, not for release profiles. In other words, there is no difference with Python Polars.
Simply deleting the test would not be a problem.

eitsupi (changes requested):

etiennebacher:
FYI I won't have time this weekend so feel free to edit this branch

tests/testthat/test-expr-expr.R
etiennebacher: > It seems that these are only panic for dev profiles, not for release profiles

This sounds to me it's unexpected and should be solved, instead of deleting the tests. Did you reproduce the same panick in dev profile with py-polars?

eitsupi: > This sounds to me it's unexpected and should be solved, instead of deleting the tests. Did you reproduce the same panick in dev profile with py-polars?

I haven't tried that, but building the R package with the release profile does not cause panic.
There is no way to know during testing which profile it was built with, and since the test will most likely be run in the dev profile, there is no point in testing.

etiennebacher: But the question is why does it fail only in the dev profile? Do you have an idea about that? I'd like to keep those tests to ensure the behavior doesn't change later on (either by choice or by mistake from upstream polars)

eitsupi: If you really want to test it, I suggest you create a reproducible example using only Rust and not from Python or R to see how it works with the dev and release profiles. Simply removing the test would suffice as I don't think there is that much value in this.

etiennebacher: Tracked here: https://github.com/pola-rs/polars/issues/19692

R/expr-expr.R
eitsupi: Done.

R/expr-expr.R
```diff
@@ -342,14 +361,16 @@ expr__gt <- function(other) {
...
-#' pl$lit(2) >= 2
-#' pl$lit(2) >= pl$lit(2)
-#' pl$lit(2)$gt_eq(pl$lit(2))
-expr__gt_eq <- function(other) {
+#' df <- pl$DataFrame(x = 1:3)
+#' df$with_columns( 
+#'   with_ge = pl$col("x")$ge(pl$lit(2)),
+#'   with_symbol = pl$col("x") >= pl$lit(2)
+#' )
+expr__ge <- function(other) {
```
etiennebacher, line 373: `gt_eq()` exists in rust-polars but not in the py-polars API (also more `ne`, `ne_missing`, `lt_eq`). Same thing as for the div operations, I think we should follow the py-polars API.

etiennebacher:
@eitsupi just for info, I don't plan on spending more time on this PR (unless you have additional comments of course). I think we can address the various TODOs at a later stage and in smaller PRs. I know this one is massive, no pressure for the review.

eitsupi (commented):
Thanks. I don't have time to review this huge change, so I will merge it anyway.
Please avoid huge PRs if you can (they make reviewing on GitHub very painful and I won't be able to merge them forever).

eitsupi:
Sorry, I forgot to correct the title and re-committed. 5b38015

R/expr-expr.R
```diff
@@ -427,16 +1274,3318 @@ expr__reshape <- function(dimensions) {
...
+#' \deqn{\alpha = 1 - \exp \left\{ \frac{ -\ln(2) }{ \lambda } \right\} \;
+#' \forall \; \lambda > 0}
+#' @param alpha Specify smoothing factor alpha directly, \eqn{0 < \alpha
+#' \leq 1}.
+#' @param adjust Divide by decaying adjustment factor in beginning periods to
+#' account for imbalance in relative weightings:
+#' * when `TRUE` (default), the EW function is calculated using weights
+#'  \eqn{w_i = (1 - \alpha)^i};
+#' * when `FALSE`, the EW function is calculated recursively by \deqn{y_0 &= x_0
+#'  \\y_t &= (1 - \alpha)y_{t - 1} + \alpha x_t}
```
eitsupi, line 3326: This line seems broken.

PR #21, opened 2024-10-18
feat: add missing list expr namespace methods and tweaks

etiennebacher:


R/expr-list.R
```diff
@@ -460,9 +475,10 @@ expr_list_head <- function(n = 5L) {
 #'   tail_by_lit = pl$col("s")$list$tail(2)
 #' )
 expr_list_tail <- function(n = 5L) {
-  offset <- -as_polars_expr(n)
-  self$`_rexpr`$list_slice(offset$`_rexpr`, as_polars_expr(n)$`_rexpr`) |>
-    wrap()
+  wrap({
+    offset <- -as_polars_expr(n)
```
eitsupi, line 479: Please use the `check_number_decimal` function to make sure that n is numeric value before that.
Like this:
https://github.com/eitsupi/neo-r-polars/blob/ec7b1e9c8ae609a14bc3ef0ab819c5254e034aee/R/dataframe-frame.R#L178

eitsupi (changes requested):

R/expr-list.R
etiennebacher: But `n` can be a string (see in the examples above) so this check cannot be done.

eitsupi: Thanks, you are right. I was mistaken.

eitsupi (approved):
Thanks for checking the details. Looks good, so let's merge it.

Issue #11, opened 2024-10-14
TODO list for error messages


etiennebacher:
1. Repetition of some calls

"Evaluation failed in `$select()`." is repeated, and I couldn't find a way to fix it:

```
pl$DataFrame(x = 1)$select(pl$col("x")$name$suffix(1))
Error:
! Evaluation failed in `$select()`.
Caused by error:
! Evaluation failed in `$select()`.
Caused by error in `pl$col("x")$name$suffix()`:
! Evaluation failed in `$suffix()`.
Caused by error:
! Argument `suffix` must be character, not double
```

2. Some calls appear in the trace but not in the user code, which is confusing

`$collect()` is not called explicitly:

```
pl$DataFrame(x = 1)$select(pl$col("a"))
Error:
! Evaluation failed in `$select()`.
Caused by error:
! Evaluation failed in `$collect()`.
Caused by error:
! ColumnNotFound(ErrString("a\n\nResolved plan until failure:\n\n\t---> FAILED HERE RESOLVING 'select' <---\nDF [\"x\"]; PROJECT */1 COLUMNS; SELECTION: None"))
```

3. Need a way to cleanup raw messages coming from rust-polars

Same code as above, the raw message about "ColumnNotFound" should be cleaned and the formatting should be applied (e.g. apply line breaks instead of `\n`). This can be done with regex but it seems a bit fragile? For instance this code:

```r
[...]
 else {
    "Evaluation failed."
 }
 cnd <- translate_polars_errors(cnd)
 abort(
   msg_body,
   call = call,
   parent = cnd
 )
}

translate_polars_errors <- function(cnd) {
  msg <- cnd$message
  new_msg <- gsub("ErrString\\((.*)\\)", "\\1", msg)

  if (startsWith(new_msg, "ColumnNotFound")) {
    new_msg <- new_msg |>
      gsub("ColumnNotFound\\(", "", x = _) |>
      gsub("Resolved .*", "", x = _) |>
      gsub("\\\\n", "", x = _) |>
      gsub("\"", "", x = _)
    new_msg <- paste("Column(s) not found:", toString(new_msg))
  }

  cnd$message <- new_msg
  cnd
}
```

gives this:

```
pl$DataFrame(x = 1)$select(pl$col("a"))
Error:
! Evaluation failed in `$select()`.
Caused by error:
! Evaluation failed in `$collect()`.
Caused by error:
! Column(s) not found: a
```

eitsupi:
[...]

PR #22, opened 2024-10-20
feat: improve error messages

etiennebacher:
Part of #11

@eitsupi just a quick PoC to see if you're fine with that, and if so I can extend this later.

Before:

```r
> pl$DataFrame(x = 1)$select(pl$col("a"))
Error:
Caused by error:
Caused by error:
! ColumnNotFound(ErrString("a\n\nResolved plan until failure:\n\n\t---> FAILED HERE RESOLVING 'select' <---\nDF [\"x\"]; PROJECT */1 COLUMNS; SELECTION: None"))
```

After:

```r
> pl$DataFrame(x = 1)$select(pl$col("a"))
Error:
Caused by error:
Caused by error:
! Column(s) not found: a

Resolved plan until failure:

	---> FAILED HERE RESOLVING 'select' <---
DF ["x"]; PROJECT */1 COLUMNS; SELECTION: None
```

---

Edit: after trimming unneessary stuff there is actually not much to review


eitsupi:
Thanks for taking a look at this. I think the direction is right.

etiennebacher:
I implemented 3 types of errors, but I'm not sure the rest are very useful for now. Their error messages already show the useful information and they seem to be mostly there because python cares more about classed errors than R does. For instance, the `ShapeMismatch` errors in snapshot already say "lengths don't match".

eitsupi (approved):
Great, thanks!

PR #26, opened 2024-10-28
feat!: add missing expr and docs to `meta` subnamespace

etiennebacher:


R/expr-meta.R
```diff
@@ -34,14 +62,22 @@ expr_meta_output_name <- function(..., raise_if_undetermined = TRUE) {
       tryCatch(
         self$`_rexpr`$meta_output_name(),
         error = function(e) {
-          NULL
+          NA_character_
```
eitsupi, line 65: Why this change?

R/import-standalone-lifecycle.R
```diff
@@ -107,8 +107,7 @@ deprecate_soft <- function(msg,
   id <- paste(id, collapse = "\n")
   verbosity <- .rlang_lifecycle_verbosity()
 
-  invisible(switch(
-    verbosity,
+  invisible(switch(verbosity,
```
eitsupi, line 110: Do not format this auto generated file.

eitsupi (changes requested):

R/expr-meta.R
etiennebacher: Because it returns `None` in python, which we usually replace by `NA` throughout the docs so I think it makes more sense (it's also what we have in the current r-polars). I can change the call to is.null() if you're ok with that.

tests/testthat/test-expr-meta.R
```diff
@@ -0,0 +1,130 @@
...
+})
+
+test_that("meta$pop", {
+  e1 <- pl$lit(40) + 2
+  e2 <- pl$lit(42)$sum()
+
+  pop1 <- e1$meta$pop()
+  expect_true(length(pop1) == 2L)
+  expect_true(pop1[[1]]$meta$eq(2))
+  expect_true(pop1[[2]]$meta$eq(40))
```
etiennebacher, line 30: @eitsupi Those fail because I don't know how to return a "proper" polars expression from a Rust function. I tried wrapping the output in `.savvy_wrap_PlRExpr` but apparently it's not enough. Could you take a look?

R/expr-meta.R
```diff
@@ -67,16 +103,177 @@ expr_meta__as_selector <- function() {
...
+#'
+#' @inherit as_polars_expr return
+#' @examples
+#' e <- pl$col("foo")$alias("bar")
+#' first <- e$meta$pop()
+#' first$meta
+expr_meta_pop <- function() {
+  wrap({
+    out <- self$`_rexpr`$meta_pop()
+    lapply(out, .savvy_wrap_PlRExpr)
```
eitsupi, line 277: Since this is a list, I think we need to `wrap` the individual elements.
Like these:
https://github.com/eitsupi/neo-r-polars/blob/fd7922069bd0512b78f4c969d62f338a280af958/R/datatypes-classes.R#L115-L121
https://github.com/eitsupi/neo-r-polars/blob/fd7922069bd0512b78f4c969d62f338a280af958/R/dataframe-frame.R#L216-L219

R/expr-meta.R
eitsupi: I see. Please update the following to use `is_na(expr_name)`.
https://github.com/eitsupi/neo-r-polars/blob/d72e64b92871a2a9b0879f0ec35025dae6c314d5/R/utils-parse-expr.R#L11-L14

eitsupi (changes requested):

R/expr-meta.R
etiennebacher: That worked, thanks

R/utils-parse-expr.R
```diff
@@ -9,7 +9,7 @@ parse_into_list_of_expressions <- function(..., `__structify` = FALSE) {
   unaliased_expr <- expr$meta$undo_aliases()
   if (unaliased_expr$meta$has_multiple_outputs()) {
     expr_name <- expr$meta$output_name(raise_if_undetermined = FALSE)
-    if (is.null(expr_name)) {
+    if (is.na(expr_name)) {
```
eitsupi, line 12:
```suggestion
    if (is_na(expr_name)) {
```

R/expr-meta.R
```diff
@@ -67,16 +103,184 @@ expr_meta__as_selector <- function() {
...
+#' pop[[1]]$meta$eq(pl$col("foo"))
+#' pop[[1]]$meta$eq(pl$col("foo"))
+expr_meta_pop <- function() {
+  wrap({
+    out <- self$`_rexpr`$meta_pop()
+    lapply(out, \(x) {
+      x |>
+        .savvy_wrap_PlRExpr() |>
+        wrap()
+    })
```
eitsupi, line 284:
```suggestion
    lapply(self$`_rexpr`$meta_pop(), \(ptr) {
        .savvy_wrap_PlRExpr(ptr) |>
          wrap()
```

eitsupi (changes requested):

R/expr-meta.R
```diff
@@ -67,16 +103,180 @@ expr_meta__as_selector <- function() {
...
+#' my_expr$meta$tree_format()
+expr_meta_tree_format <- function(..., return_as_string = FALSE) {
+  wrap({
+    check_dots_empty0(...)
+    out <- self$`_rexpr`$meta_tree_format()
+    if (isTRUE(return_as_string)) {
+      out
+    } else {
+      cat(out)
+    }
```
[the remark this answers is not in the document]
etiennebacher: I'm not convinced about that, that's precisely what this argument is about. It would be counter-intuitive to have `return_as_string = FALSE` but still return a string (even if it's not printed in the console).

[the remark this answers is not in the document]
eitsupi: Hmmm, I don't think the API of this function is so beautiful. Why do we need to specify a long argument like `return_as_string=TRUE` to get the output?
If the function returns a string from the beginning, we can just connect it to `cat()`.

There is a similar function `<LazyFrame>$explain()`, which always returns a string.
https://github.com/eitsupi/neo-r-polars/blob/c7461be38ac208d1275c8bdf78a2c2be6d896367/R/lazyframe-frame.R#L188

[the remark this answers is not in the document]
etiennebacher: I don't mind removing all the args and just return a string, I'll change that

R/expr-meta.R
```diff
@@ -188,23 +188,14 @@ expr_meta_root_names <- function() {
 
 #' Format the expression as a tree
 #'
-#' @inheritParams rlang::check_dots_empty0
-#' @param return_as_string Return the tree as a character vector? If `FALSE`
-#' (default), the tree is printed in the console.
-#'
 #' @inherit as_polars_expr return
```
eitsupi, line 191: return should be a character vector.

eitsupi (approved):
Thanks!
Let's merge them.

etiennebacher:
Thanks for the review!

PR #27, opened 2024-10-30
!! DO NOT MERGE !! Use `"use-custom-error"` feature of savvy

yutannihilation:
I'm planning to provide `impl<E: std::error::Error> From<E> for savvy::Error` for easy error-conversion like anyhow.

Details: https://github.com/yutannihilation/savvy/pull/324

This is a breaking change in that this will conflicts with `impl From<RPolarsErr> for savvy::Error`. So, savvy will provide a feature `use-custom-error` to opt-out the conversion of `From<dyn std::error::Error>`. This pull request is to see if it works. Please let me know if you have any concerns.

```
   error[E0119]: conflicting implementations of trait `From<RPolarsErr>` for type `savvy::Error`
     --> src\error.rs:27:1
      |
   27 | impl From<RPolarsErr> for savvy::Error {
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: conflicting implementation in crate `savvy`:
              - impl<E> From<E> for savvy::Error
                where E: StdError, E: 'static;
```

### DIFF
--- a/src/rust/Cargo.lock
+++ b/src/rust/Cargo.lock
@@ -2527,8 +2527,7 @@ dependencies = [
 [[package]]
 name = "savvy"
 version = "0.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9901422223436007d21f5448a8125b05c2243ede8f48fe5701349ff6c59a43ad"
+source = "git+https://github.com/yutannihilation/savvy.git?branch=feat/error-conversion2#10512b3b26955edad412d6e25c0a3f159dce1038"
 dependencies = [
  "cc",
  "once_cell",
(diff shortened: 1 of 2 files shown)


eitsupi (commented):
Thanks for taking a look at this!

My question is, do we really need this feature here?
Would simply implementing something like `impl From<PolarsError> for RPolarsErr` work without this feature?
(although of course you don't need to do that with this PR)

yutannihilation:
In my understanding, you anyway need to convert `RPolarsErr` to `savvy::Error`. If you can remove either of the following impls, it should compile without `use-custom-error` feature.

https://github.com/eitsupi/neo-r-polars/blob/ab56d538c9c475c6f43d7bf756ecc1228ae8e7f6/src/rust/src/error.rs#L7-L8
https://github.com/eitsupi/neo-r-polars/blob/ab56d538c9c475c6f43d7bf756ecc1228ae8e7f6/src/rust/src/error.rs#L27

PR #28, opened 2024-10-30
feat: add `_horizontal` functions

etiennebacher:


R/functions-aggregation-horizontal.R
```diff
@@ -0,0 +1,145 @@
...
+#'   c = c("x", "y", "z")
+#' )
+#' df$with_columns(
+#'   min = pl$min_horizontal("a", "b")
+#' )
+pl__min_horizontal <- function(...) {
+  wrap({
+    check_dots_unnamed()
+    by <- parse_into_list_of_expressions(...)
+    min_horizontal(by)
```
eitsupi, line 22: How about?

```suggestion
    parse_into_list_of_expressions(...) |>
      min_horizontal()
```

src/rust/src/functions/aggregation.rs
```diff
@@ -0,0 +1,46 @@
+use crate::{prelude::*, PlRExpr, RPolarsErr};
+use polars::lazy::dsl;
+use polars::prelude::Expr;
```
eitsupi, line 3: I don't think this is needed.

```suggestion
```

eitsupi (approved):
Thanks!
A few minor comments.

PR #30, opened 2024-10-31
feat: add `pl$scan_csv()` and `pl$read_csv()`

etiennebacher:


etiennebacher:
@eitsupi when you have time, it would be great if you could review this one first so that I can reuse it to implement other readers

R/io-csv.R
```diff
@@ -0,0 +1,284 @@
...
+    storage_options = NULL,
+    # credential_provider = NULL,
+    retries = 2,
+    file_cache_ttl = NULL,
+    include_file_paths = NULL) {
+  wrap({
+    check_dots_empty0(...)
+    if (length(source) == 0) {
+      abort("`source` must have length > 0.")
+    }
```
eitsupi, line 154: Don't we need to check the type of `source` in advance?

R/io-csv.R
```diff
@@ -0,0 +1,284 @@
...
+      null_values <- as.list(null_values)
+    } else if (!is.null(null_values)) {
+      abort("`null_values` must be a character vector or a named list.")
+    }
+
+    schema_overrides <- lapply(schema_overrides, \(x) x$`_dt`)
+
+    if (length(schema) > 0) {
+      schema <- lapply(schema, \(x) x$`_dt`)
+    }
```
eitsupi, line 185: Do you want to use `parse_into_list_of_datatypes`?

R/io-csv.R
etiennebacher: Done

R/io-csv.R
etiennebacher: Thanks, forgot about that one

R/io-csv.R
```diff
@@ -0,0 +1,282 @@
...
+    if (is.null(row_index_name) && !is.null(row_index_offset)) {
+      row_index_offset <- NULL
+    }
+
+    if (!is.null(storage_options)) {
+      all_character <- all(vapply(storage_options, is_character, n = 1, FUN.VALUE = logical(1)))
+      if (!all_character) {
+        abort("All elements of `storage_options` must be of type character.")
+      }
+    }
```
eitsupi, line 193: Ditto.
My understanding is that a named vector is sufficient for this.

Something like: https://github.com/eitsupi/r-glaredb/blob/d011a337187e42421c7db08b75515599ba7651be/src/rust/src/connect.rs#L52-L71

R/io-csv.R
eitsupi, line 1: Rename the file to `io-csv-functions.R` to match the name with Python Polars.

eitsupi (changes requested):
Sorry for the late review. Honestly, CSV is so complex that I think it would be better to implement a simple reader like Arrow first.

R/io-csv.R
etiennebacher: Done

R/io-csv-functions.R
```diff
@@ -0,0 +1,253 @@
...
+  wrap({
+    check_dots_empty0(...)
+    check_character(source)
+    if (length(source) == 0) {
+      abort("`source` must have length > 0.")
+    }
+    check_list_of_polars_dtype(schema_overrides, allow_null = TRUE)
+    check_list_of_polars_dtype(schema, allow_null = TRUE)
+    check_named_list(schema_overrides, allow_null = TRUE)
+    check_named_list(schema, allow_null = TRUE)
```
etiennebacher, line 156: Do you think it's worth adding a `check_has_names` param in `check_list_of_polars_dtype()` instead of having 2 functions per param?

eitsupi: In my opinion, it is not a good idea to enforce names with something like a schema.
The limitation of not being able to use empty string in the name prevents things like the following from working (It works on the branch main for now):

```r
pl$DataFrame(1:5, foo = letters[1:5]) |>
  as.data.frame() |>
  arrow::write_dataset("data", format = "arrow", partitioning = "")
pl$read_ipc("data", hive_schema = list(pl$UInt32))
```

R/io-csv-functions.R
```diff
@@ -0,0 +1,251 @@
...
+    truncate_ragged_lines = FALSE,
+    decimal_comma = FALSE,
+    glob = TRUE,
+    storage_options = NULL,
+    retries = 2,
+    file_cache_ttl = NULL,
+    include_file_paths = NULL) {
+  wrap({
+    check_dots_empty0(...)
+    check_character(source)
```
eitsupi, line 149:
```suggestion
    check_character(source, allow_na = FALSE)
```

src/rust/src/lazyframe/general.rs
```diff
@@ -273,9 +270,149 @@ impl PlRLazyFrame {
...
+                Some(vec)
+            }
+            None => None,
+        };
+        let overwrite_dtype = overwrite_dtype.map(|overwrite_dtype| {
+            overwrite_dtype
+                .into_iter()
+                .map(|(name, dtype)| Field::new((&*name).into(), dtype))
+                .collect::<Schema>()
+        });
```
eitsupi, line 395: `<Wrap<Vec<Field>>>::try_from(overwrite_dtype)` ?

eitsupi (changes requested):

src/rust/src/lazyframe/general.rs
etiennebacher: I don't understand what you want to replace, can you apply your change?

eitsupi (approved):
Let's merge for now.

PR #31, opened 2024-11-05
feat: add missing functions and docs for LazyFrame

etiennebacher:


etiennebacher:
@eitsupi I start testing those functions and I find it a bit hard to read this kind of tests:

```r
expect_query_equal(
    .input$select("int32"),
    .data,
    pl$DataFrame(int32 = 1:5)
  )
```

I think it's a good idea to test LazyFrame and DataFrame functions at once instead of manually copying tests, but the function above cannot take multiple datasets when testing joins for example. In `tidypolars` I have a simple script that copies test files for DataFrame and replace all occurrences of DataFrame by LazyFrame and more: https://github.com/etiennebacher/tidypolars/blob/main/tests/testthat/setup.R

This way, I can just write tests for DataFrame and it automatically generates the same tests for LazyFrame. It's also easier to see if an error only happens for eager/lazy data since tests appear in different files. What do you think about this approach?

eitsupi:
Sorry I did not notice the mention.

> What do you think about this approach?

I think it's a good idea, but it may be difficult given the support for tests for things like head that behave differently in LazyFrame and DataFrame?

I'm also concerned about the possibility of files being updated on CI if run as part of `R CMD check` like `setup.R`, but I think that's a minor issue.

> the function above cannot take multiple datasets when testing joins for example.

Could this be addressed by updating the function?

etiennebacher:
> I think it's a good idea, but it may be difficult given the support for tests for things like head that behave differently in LazyFrame and DataFrame?

I've done this change in 7a87e3deebc4977ebefe77f933d57d00c324467f.

Basically, we can write tests for DataFrames in `test-frame.R`. When `devtools::test()` runs, this file is duplicated to `test-frame-lazy.R`, where all functions related to DataFrames are replaced (e.g. `as_polars_df()` -> `as_polars_lf()`, etc.). 

As you said, some functions behave differently between eager and lazy (e.g. `slice()`). When this is the case, we can branch the tests depending on the data, for instance:
```r
  if (is_in_lazy_test()) {
    expect_error(
      .data$slice(0, -2),
      r"(-2.0 is out of range that can be safely converted to u32)"
    )
  } else {
    expect_equal(
      .data$slice(0, -2),
      pl$DataFrame(foo = 1:3, bar = 6:8)
    )
  }
```
When `test-frame-lazy.R` runs, the first branch is tested, and when `test-frame.R` runs the second branch is tested. Also, if a function doesn't have an equivalent at all in lazy (e.g. `to_struct()`) then we can use `skip_if(is_in_lazy_test())`.
 
This takes slightly more lines than the previous approach but it is easier to read and to customize IMO.

> I'm also concerned about the possibility of files being updated on CI if run as part of R CMD check like setup.R

I don't understand this comment. We simply copy the content of `test-frame.R` to a new file and automatically replace parts of its content. Whether or not it runs in CI doesn't change the content of the test files and I've never had an issue like that in `tidypolars`.

Let me know what you think before I add more tests.

tests/testthat/test-frame.R
eitsupi, line 1: Please rename the file as it will not be possible to specify only this file when filtering with the `devtools::test` function.

tests/testthat/test-frame-lazy.R
```diff
@@ -0,0 +1,206 @@
...
+    )
+  } else {
+    expect_equal_lazy(
+      .data$tail(-4),
+      pl$LazyFrame(foo = 5L, bar = 10L)
+    )
+  }
+})
+
+Sys.setenv(POLARS_IN_LAZY_TEST = FALSE)
```
eitsupi, line 206: Is there any reason not to use withr functions to change environment variables?

eitsupi (changes requested):
My point is that it is possible to pass the CI even if the file is not updated. In my opinion file overwriting should not be part of the R CMD test.

I don't like this approach because it makes it impossible to tell the correspondence between the file in which the function is written and the file in which the test is written.
Basically, the methods that exist in LazyFrame are dispatched to DataFrame, so it seems more natural to automatically test the tests written for LazyFrame in DataFrame as well.

I don't think most of the DataFrame methods are inherently worth testing, just as the methods of Expr that are dispatched to Series are not tested one by one.
(For convenience, we have written functions for DataFrame manually, but it should even be possible to automatically dispatch them as we automatically dispatch the methods for Expr in Series to save time and effort in updating them.)

In other words, if we adopt the replacement approach, we would need to test for DataFrame on the auto-generated file that replaced `test-lazyframe-frame.R`. Separately, DataFrame-specific tests should be performed on the `test-dataframe-frame.R`.

tests/testthat/test-frame.R
etiennebacher: I don't understand, one can still to `devtools::test("tests/testthat/test-frame.R")`, no?

I have named the files `test-frame` and `test-frame-lazy` so that they are next to each other in the file tree. Also, why do you prefer `test-lazyframe-frame` instead of `test-lazyframe`?

etiennebacher:
> I don't think most of the DataFrame methods are inherently worth testing, just as the methods of Expr that are dispatched to Series are not tested one by one.

I'd prefer to test them too. If we only test for lazy and later the behavior for eager becomes different from the behavior for lazy we won't know.

---

I went back to the previous custom expectations and did some changes in [`81b954c` (#31)](https://github.com/eitsupi/neo-r-polars/pull/31/commits/81b954c37bd8ef721d2608bf27000d68af0080b4). It is now possible to pass queries with multiple inputs, which is necessary for joins for instance:
```r
test_that("joins work lazy/eager", {
  df <- pl$DataFrame(
    foo = 1:3,
    bar = c(6, 7, 8),
    ham = c("a", "b", "c")
  )

  other_df <- pl$DataFrame(
    apple = c("x", "y", "z"),
    ham = c("a", "b", "d")
  )

  expect_query_equal(
    .input$join(.input2, on = "ham"),
    .input = df, .input2 = other_df,
    pl$DataFrame(
      foo = 1:2,
      bar = c(6, 7),
      ham = c("a", "b"),
      apple = c("x", "y")
    )
  )
})
```

R/utils-various.R
```diff
@@ -16,3 +16,116 @@ extend_bool <- function(value, n_match, value_name, match_name) {
...
+    if (length(invalid) > 0) {
+      msg <- paste0("`", invalid, "`", collapse = ", ")
+      abort(
+        paste0("In `statistics`,", msg, "are not valid keys."),
+        call = call
+      )
+    }
+  }
+  statistics
+}
```
eitsupi, line 131: Do these functions exist in Python Polars? If they do not exist, please do not add them here to separate them into a different PR. I do not want to see them here as they are too specific.

man/IO_sink_csv.Rd
eitsupi, line 1: Please do not set rdname tag.

eitsupi (changes requested):
Please resolve conflicts.

src/rust/src/lazyframe/general.rs
```diff
@@ -53,7 +52,7 @@ impl PlRLazyFrame {
         cluster_with_columns: bool,
         streaming: bool,
         _eager: bool,
-    ) -> Result<Self> {
+    ) -> Result<PlRLazyFrame> {
```
eitsupi, line 55: Why?

eitsupi (changes requested):

R/lazyframe-frame.R
```diff
@@ -45,7 +45,6 @@ wrap.PlRLazyFrame <- function(x, ...) {
   self
 }
 
-# TODO: link to pl__select
```
eitsupi, line 48: Why?

R/lazyframe-frame.R
```diff
@@ -254,33 +519,98 @@ lazyframe__cast <- function(..., .strict = TRUE) {
...
+  descending <- extend_bool(descending, length(by), "descending", "...")
+  nulls_last <- extend_bool(nulls_last, length(by), "nulls_last", "...")
+
+  self$`_ldf`$sort_by_exprs(
+    by,
+    descending = descending,
+    nulls_last = nulls_last,
+    multithreaded = multithreaded,
+    maintain_order = maintain_order
+  ) |> wrap()
```
eitsupi, line 613: Why?

R/lazyframe-frame.R
```diff
@@ -332,20 +662,98 @@ lazyframe__sort <- function(
 #'   })
 #' }
 lazyframe__with_columns <- function(...) {
+  structify <- parse_env_auto_structify()
+
+  parse_into_list_of_expressions(..., `__structify` = structify) |>
+    self$`_ldf`$with_columns() |>
+    wrap()
+}
```
eitsupi, line 670: Why

R/lazyframe-frame.R
```diff
@@ -332,20 +662,98 @@ lazyframe__sort <- function(
...
+#'   foo = 1:3,
+#'   bar = c(6, 7, 8),
+#'   ham = c("a", "b", "c")
+#' )
+#' lf$drop("ham")$collect()
+#' lf$drop("ham", "bar")$collect()
+#'
+#' # Drop multiple columns by passing a selector
+#' lf$drop(cs$all())$collect()
+lazyframe__drop <- function(..., .strict = TRUE) {
```
eitsupi, line 752: Why?
Because `...` are unnamed, we don't need to use `.strinct` here.

R/lazyframe-frame.R
```diff
@@ -185,6 +255,176 @@ lazyframe__collect <- function(
...
+#'   pl$col("bar")$cast(pl$Int64)
+#' )$collect_schema()
+lazyframe__collect_schema <- function() {
+  wrap({
+    lapply(self$`_ldf`$collect_schema(), function(x) {
+      .savvy_wrap_PlRDataType(x) |>
+        wrap()
+    })
+  })
+}
```
eitsupi, line 285: collect_schema is already added in the main branch. Please remove this one.

eitsupi (changes requested):

man/IO_sink_csv.Rd
etiennebacher: This is because on the website we gather those IO functions together and it's easier to select all Rd files starting with "IO". The user can still access the docs with `?lazyframe__sink_csv`, it doesn't change that.

R/lazyframe-frame.R
etiennebacher: I don't see how this should be mentioned in the docs of `$select()` but I can add something if you have a suggestion

R/lazyframe-frame.R
etiennebacher: I got confused at some point on whether `wrap()` should include the whole code or just the call to `self$`_ldf`$<fn>()`, I'll revert this.

R/lazyframe-frame.R
etiennebacher: I didn't change this code, it's just git showing diffs in a weird way (the code is almost the same for `with_columns_seq()`, hence the confusion I guess)

R/utils-various.R
etiennebacher: `translate_statistics()` doesn't exist but we need more input validation than in python (e.g. in py-polars I think pyo3 takes care of checking of invalid keys but here we do it in R).

For `make_profile_plot()` they instead put all the plot-related code in the definition of `profile()`. I don't mind moving all this code in the body of `profile()` but it's cleaner to have this in a util function since it only matters if `show_plot = TRUE`.

man/IO_sink_csv.Rd
eitsupi: In my opinion classification of Rd files by filename should not be done as I found it very difficult with the current r-polars, I prefer to index them manually as per the Python Polars API documentation.
So please remove them.

R/lazyframe-frame.R
eitsupi: My point is why are you deleting TODOs for me without my consent. Please do not delete them.

src/rust/src/lazyframe/general.rs
eitsupi: @etiennebacher Your change covers all functions, so don't just fix one and mark it as resolved.
Why did you change everything?

foo.parquet
eitsupi, line 1: Why was this file committed?

src/rust/src/conversion/mod.rs
```diff
@@ -578,7 +611,7 @@ impl TryFrom<&str> for Wrap<Roll> {
             "raise" => Roll::Raise,
             "forward" => Roll::Forward,
             "backward" => Roll::Backward,
-            _ => return Err("unreachable".to_string()),
+            _ => return Err(format!("unreachable",)),
```
eitsupi, line 614: Why?

src/rust/src/conversion/mod.rs
```diff
@@ -601,6 +634,196 @@ impl TryFrom<&str> for Wrap<QuantileMethod> {
...
+        "lz4" => ParquetCompression::Lz4Raw,
+        "zstd" => ParquetCompression::Zstd(
+            compression_level
+                .map(|lvl| {
+                    ZstdLevel::try_new(lvl)
+                        .map_err(|e| savvy::Error::new(format!("{e:?}").as_str()))
+                })
+                .transpose()?,
+        ),
+        _ => unreachable!(),
```
eitsupi, line 752: Should not panic.

eitsupi (changes requested):

src/rust/src/conversion/mod.rs
etiennebacher: Happened at some point when merging with main I guess

src/rust/src/lazyframe/general.rs
etiennebacher: Got confused as in the codebase sometimes we specify the output type instead of `Self`, reverted

man/IO_sink_csv.Rd
etiennebacher: I've made the change but I don't get your point. Apart from the scripts building the website, when do you deal with Rd files directly?

R/dataframe-frame.R
```diff
@@ -539,3 +539,26 @@ dataframe__rechunk <- function() {
...
+#'   select(
+#'   pl$struct("b"),
+#'   a_and_c = pl$struct(c("a", "c"))
+#' )
+#' df
+#'
+#' df$unnest("a_and_c")
+dataframe__unnest <- function(...) {
+  wrap({
+    columns <- unlist(list2(...))
```
eitsupi, line 560: `unlist()` is dangerous.
Please use this as like `pl__col()`

https://github.com/eitsupi/neo-r-polars/blob/473ade21c048bd4a80cec7b201dfb978f5c5b5c1/R/utils-check.R#L34-L62

R/io-csv-functions.R
```diff
@@ -112,38 +110,39 @@
...
+  eol_char = "\n",
+  raise_if_empty = TRUE,
+  truncate_ragged_lines = FALSE,
+  decimal_comma = FALSE,
+  glob = TRUE,
+  storage_options = NULL,
+  retries = 2,
+  file_cache_ttl = NULL,
+  include_file_paths = NULL
+) {
```
eitsupi, line 145: Why?

R/io-csv-functions.R
```diff
@@ -210,38 +209,39 @@ pl__scan_csv <- function(
...
```
eitsupi, line 244: Why?

src/rust/src/conversion/mod.rs
```diff
@@ -601,6 +634,196 @@ impl TryFrom<&str> for Wrap<QuantileMethod> {
...
+                        let tmp = val.to_vec();
+                        *tmp.first().unwrap()
+                    }
+                    _ => unreachable!(),
+                };
+                (name, value)
+            })
+            .collect::<std::collections::HashMap<&str, bool>>();
+        let mut out = StatisticsOptions::default();
+        out.min_value = *hm.get(&"min").unwrap();
```
eitsupi, line 777: Should not use `unwrap()`

eitsupi (changes requested):

R/io-csv-functions.R
etiennebacher: Looks like `styler` dev version changes indentation, reverted for now

R/dataframe-frame.R
etiennebacher: Should be good now, but note that there's a different behavior between eager and lazy: https://github.com/pola-rs/polars/issues/20128

src/rust/src/conversion/mod.rs
etiennebacher: We already do input validation on the R side so it's not clear to me how to remove the `unwrap()` without duplicating the effort and having too verbose code. Can you help me out on this one?

man/IO_sink_csv.Rd
eitsupi: I am not sure what your question means.
I just imagine that the index of the current Python Polars documentation has been manually updated.
No automatic classification by Rd file name is performed, so there is no need to set the Rd file name to something special.

R/lazyframe-frame.R
```diff
@@ -96,41 +96,112 @@ wrap.PlRLazyFrame <- function(x, ...) {
 lazyframe__select <- function(...) {
   wrap({
     structify <- parse_env_auto_structify()
-
```
eitsupi, line 99: Why?

R/lazyframe-frame.R
```diff
@@ -57,12 +57,12 @@ wrap.PlRLazyFrame <- function(x, ...) {
 #' `$select()` call. For instance, if you create a variable `x`, you will only
 #' be able to use it in another `$select()` or `$with_columns()` call.
 #'
-#' @inherit pl__LazyFrame return
-#' @param ... <[`dynamic-dots`][rlang::dyn-dots]>
-#' Name-value pairs of objects to be converted to polars [expressions][Expr]
-#' by the [as_polars_expr()] function.
-#' Characters are parsed as column names, other non-expression inputs are parsed as [literals][pl__lit].
-#' Each name will be used as the expression name.
+#' @inherit as_polars_lf return
```
eitsupi, line 60: Why?

R/lazyframe-frame.R
```diff
@@ -57,12 +57,12 @@ wrap.PlRLazyFrame <- function(x, ...) {
...
+#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> Name-value pairs of objects
+#' to be converted to polars [expressions][Expr] by the [as_polars_expr()]
+#' function. Characters are parsed as column names, other non-expression inputs
+#' are parsed as [literals][pl__lit]. Each name will be used as the expression
+#' name.
```
eitsupi, line 65: Why?

R/lazyframe-frame.R
```diff
@@ -96,41 +96,112 @@ wrap.PlRLazyFrame <- function(x, ...) {
...
+#' @param streaming `r lifecycle::badge("experimental")` Logical. Process the
+#' query in batches to handle larger-than-memory data. If `FALSE` (default),
+#' the entire query is processed in a single batch.
+#' @param _eager A logical, indicates to turn off multi-node optimizations and
+#' the other optimizations. This option is intended for internal use only.
+#'
+#' @inherit as_polars_lf return
+#'
+#' @seealso
+#'  - [`$fetch()`][lazyframe__fetch] - fast limited query check
```
eitsupi, line 196: `fetch` should be removed.

R/lazyframe-frame.R
```diff
@@ -96,41 +96,112 @@ wrap.PlRLazyFrame <- function(x, ...) {
...
+#'  occur on self-joins or unions.
+#' @param comm_subexpr_elim Logical. Common subexpressions will be cached and
+#' reused.
+#' @param cluster_with_columns Combine sequential independent calls to
+#' [`with_columns()`][lazyframe__with_columns].
+#' @param streaming `r lifecycle::badge("experimental")` Logical. Process the
+#' query in batches to handle larger-than-memory data. If `FALSE` (default),
+#' the entire query is processed in a single batch.
+#' @param _eager A logical, indicates to turn off multi-node optimizations and
+#' the other optimizations. This option is intended for internal use only.
```
eitsupi, line 191: Why?

eitsupi (changes requested):
Sorry, but there are too many irrelevant changes to review them all.
Could you please review everything once and undo all the irrelevant changes?
Or split the pull request.

R/lazyframe-frame.R
```diff
@@ -358,12 +759,2052 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+    } else {
+      if (missing(mapping) || !is.list(mapping)) {
+        mapping <- list2(...)
+      }
+      existing <- names(mapping)
+      new <- unlist(mapping)
+      self$`_ldf`$rename(existing, new, strict)
+    }
+  })
+}
```
eitsupi, line 1386: At first glance, this function looks like something that should be rewritten.

R/lazyframe-frame.R
```diff
@@ -358,12 +759,2052 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#' @examples
+#' # fetch 3 rows
+#' as_polars_lf(iris)$fetch(3)
+#'
+#' # this fetch-query returns 4 rows, because we started with 3 and appended one
+#' # row in the query (see section 'Details')
+#' as_polars_lf(iris)$
+#'   select(pl$col("Species")$append("flora gigantica, alien"))$
+#'   fetch(3)
+lazyframe__fetch <- function(
```
eitsupi, line 1422:
```suggestion
lazyframe___fetch <- function(
```

R/lazyframe-frame.R
```diff
@@ -358,12 +759,2052 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'   data
+#' }
+#' df1 <- as_polars_lf(iris)
+#' df2 <- give_attr(df1)
+#'
+#' # now, the original LazyFrame doesn't get this attribute
+#' attributes(df1)
+lazyframe__clone <- function() {
+  self$`_ldf`$clone_in_rust()
+}
```
eitsupi, line 1654: Where does `clone_in_rust` come from?

eitsupi (changes requested):

tests/testthat/test-lazyframe-frame.R
```diff
@@ -129,6 +129,1269 @@ test_that("slice/head/tail works lazy/eager", {
     .input$tail(-4),
     .data,
     pl$DataFrame(foo = 5L, bar = 10L),
-    r"(-4\.0 is out of range that can be safely converted to u32)"
+    r"(-4.0 is out of range that can be safely converted to u32)"
```
eitsupi, line 132: Why?

eitsupi (changes requested):

R/dataframe-frame.R
eitsupi: Ok, please remove the dataframe method and use the lazyframe one.

R/lazyframe-frame.R
etiennebacher: This is quite similar to the python implementation: https://github.com/pola-rs/polars/blob/87feed72585eff5acf3defb7f81029123d5cba68/py-polars/polars/lazyframe/frame.py#L5274-L5279

```
        if callable(mapping):
            return self.select(F.all().name.map(mapping))
        else:
            existing = list(mapping.keys())
            new = list(mapping.values())
            return self._from_pyldf(self._ldf.rename(existing, new, strict))
```
Since we want to use `mapping` if it's provided and `...` if `mapping` is absent we're going to have something a bit more complicated. What do you think about something like this?

```r
lazyframe__rename <- function(mapping, ..., .strict = TRUE) {
  wrap({
    if (!missing(mapping)) {
      check_dots_empty0(...)
      if (is_function(mapping)) {
        self$select(pl$all()$name$map(mapping))
      } else {
        abort("`mapping` must be a function when provided.")
      }
    } else {
      # Have a check_dots_named() function?
      mapping <- list2(...)
      existing <- names(mapping)
      new <- unlist(mapping)
      self$`_ldf`$rename(existing, new, .strict)
    }
  })
}
```

R/lazyframe-frame.R
etiennebacher: I removed this function, per your comment above

R/lazyframe-frame.R
```diff
@@ -270,6 +587,9 @@ lazyframe__sort <- function(
     check_dots_unnamed()
 
     by <- parse_into_list_of_expressions(...)
+    if (length(by) == 0) {
+      abort("`...` must contain at least one element.")
+    }
```
etiennebacher, line 592: In py-polars, sort() on empty errors because it requires a positional argument but here, without this condition it returns the input.

eitsupi: Perhaps `!missing(...)` would be more appropriate?

R/lazyframe-frame.R
eitsupi: In my opinion, functions that accept Python dict, such as this `rename` and `cast`, which is already in the main branch, should use dynamic dots instead of dict.

https://github.com/eitsupi/neo-r-polars/blob/365a66965243cfeb3909f81fa576b28aa01cd102/R/dataframe-frame.R#L427-L447

In any case, since the map_name function does not yet exist, I believe it is a TODO if a function is received.

etiennebacher: > In my opinion, functions that accept Python dict, such as this `rename` and `cast`, which is already in the main branch, should use dynamic dots instead of dict.

Agreed, the only tricky part here is that it can also accept a function. What do you think of the following?

```r
lazyframe__rename <- function(..., .strict = TRUE) {
  wrap({
    mapping <- list2(...)
    if (length(mapping) == 1 && is_function(mapping)) {
      check_dots_empty0(...)
      # TODO: this requires $name$map()
      return(self$select(pl$all()$name$map(mapping)))
    }
    existing <- names(mapping)
    new <- unlist(mapping)
    self$`_ldf`$rename(existing, new, .strict)
  })
}
```
I have added a TODO meanwhile.

eitsupi: > What do you think of the following?

Thank you. I would prefer something like that. (Just wondering if a type check on the `mapping` would be necessary?)

etiennebacher: Done in [`25dcda8` (#31)](https://github.com/eitsupi/neo-r-polars/pull/31/commits/25dcda8db9289e19a970d4abf2722f35cb9d7c16). 

> Just wondering if a type check on the mapping would be necessary?

We could add something like 

```r
if (!all(vapply(mapping, is_character, logical(1)))) {
      abort("`...` only accepts a function or named characters.")
    }
```
although I'm not very happy with the wording.

eitsupi: Since this is a frequent pattern, you can probably use `is_list_of_string()` to check.

etiennebacher: Done

eitsupi (changes requested):
Could you fix tests?
Some tests seem broken.

tests/testthat/test-lazyframe-frame.R
```diff
@@ -1,4 +1,4 @@
-test_that("select works lazy/eager", {
```
eitsupi, line 1: Why?

tests/testthat/test-lazyframe-frame.R
```diff
@@ -56,13 +56,13 @@ test_that("POLARS_AUTO_STRUCTIFY works for select", {
...
         as_polars_lf(.data)$select(
           is_odd = pl$struct(((pl$col(pl$Int32) %% 2) == 1)$name$suffix("_is_odd")),
-        )$collect()
+        )
       )
     }
   )
 })
 
-test_that("slice/head/tail works lazy/eager", {
```
eitsupi, line 65: Why?

tests/testthat/test-lazyframe-frame.R
```diff
@@ -93,7 +93,7 @@ test_that("slice/head/tail works lazy/eager", {
     .input$slice(0, -2),
     .data,
     pl$DataFrame(foo = 1:3, bar = 6:8),
-    r"(negative slice length \(-2\) are invalid for LazyFrame)"
+    r"(-2.0 is out of range that can be safely converted to u32)"
```
eitsupi, line 96:
```suggestion
    r"(-2\.0 is out of range that can be safely converted to u32)"
```

tests/testthat/test-lazyframe-frame.R
```diff
@@ -56,13 +56,13 @@ test_that("POLARS_AUTO_STRUCTIFY works for select", {
         .data,
         as_polars_lf(.data)$select(
           is_odd = pl$struct(((pl$col(pl$Int32) %% 2) == 1)$name$suffix("_is_odd")),
-        )$collect()
```
eitsupi, line 59: Why?

eitsupi (changes requested):

eitsupi:
One suggestion: Why not split this pull request?
I think it should at least be split between changing the test structure and adding methods.
It seems to be a very tricky situation with both test corruption and method corruption currently occurring.

tests/testthat/test-lazyframe-frame.R
```diff
@@ -132,3 +132,1266 @@ test_that("slice/head/tail works lazy/eager", {
...
+  df <- pl$LazyFrame(
+    index = c(1:5, 6.0),
+    a = c(3, 7, 5, 9, 2, 1)
+  )
+
+  expect_grepl_error(
+    df$rolling(index_column = "index", period = "2i")$agg(
+      pl$sum("a")$alias("sum_a")
+    )$collect()
+  )
```
eitsupi, line 732: What is this function?

etiennebacher:
Yeah, it was easier to implement methods and tests at the same time for string expressions for instance, but this one is too massive, I'll split it.

PR #32, opened 2024-11-06
feat: add missing functions and docs for selectors

etiennebacher:


R/selectors-s3-base.R
eitsupi, line 1: This file is needed?
selectors are subclass of expr, so I don't think implement these are ideal.

etiennebacher: Indeed, I removed it. Nice catch

src/rust/src/functions/lazy.rs
```diff
@@ -113,6 +113,32 @@ pub fn dtype_cols(dtypes: ListSexp) -> Result<PlRExpr> {
     Ok(dsl::dtype_cols(&dtypes).into())
 }
 
+#[savvy]
+pub fn index_cols(indices: NumericSexp) -> Result<PlRExpr> {
+    let n: Vec<i64> = indices
+        .as_slice_i32()?
+        .iter()
+        .map(|x| *x as i64)
+        .collect::<Vec<_>>();
```
eitsupi, line 122: Please use this:
https://github.com/eitsupi/neo-r-polars/blob/7dea582e0afa34b7b2fc654f5d8a6477c90504aa/src/rust/src/conversion/mod.rs#L242

R/functions-col.R
```diff
@@ -64,3 +64,57 @@ pl__col <- function(...) {
...
+#'   b = c(4, 5, 2),
+#'   c = c("foo", "bar", "baz")
+#' )
+#'
+#' df$select(pl$nth(1))
+#' df$select(pl$nth(c(2, 0)))
+pl__nth <- function(indices) {
+  wrap({
+    index_cols(indices)
+  })
```
eitsupi, line 85: Please check no NA.

eitsupi (changes requested):

eitsupi (changes requested):

R/selectors.R
```diff
@@ -45,7 +59,7 @@ wrap_to_selector <- function(x, name, parameters = NULL) {
 selector__invert <- function() {
   inverted <- cs$all()$sub(self)
   # TODO: we want to print something like `!cs$all()` when call `!cs$all()`
-  inverted$`_print_override` <- deparse1(sys.call(sys.nframe() - 1))
+  # inverted$`_print_override` <- deparse1(sys.call(sys.nframe() - 1))
```
eitsupi, line 62: Why?

R/selectors.R
```diff
@@ -101,3 +376,680 @@ cs__by_name <- function(..., .require_all = TRUE) {
...
+#' # Group by all string columns, sum the numeric columns, then sort by the
+#' # string cols:
+#' df$group_by(cs$string())$agg(cs$numeric()$sum())$sort(cs$string())
+#'
+#' # Group by all string and categorical columns:
+#' df$
+#'   group_by(cs$string(include_categorical = TRUE))$
+#'   agg(cs$numeric()$sum())$
+#'   sort(cs$string(include_categorical = TRUE))
+cs__string <- function(include_categorical = FALSE) {
```
eitsupi, line 973:
```suggestion
cs__string <- function(..., include_categorical = FALSE) {
```

eitsupi (changes requested):

R/selectors.R
etiennebacher: Otherwise, this throws an error:

```r
df$select(!cs$alpha())

$<- - syntax error: Assigning to the
  member `_print_override` of this polars object
  is not supported
```

eitsupi: This is to mimic the Python Polars print results as noted in the TODO comment above.
Do you know of a way to achieve this?

etiennebacher: I tried a couple of things but no success. TBH I don't think the print method of selectors is a priority.

eitsupi: Ok, thanks.

R/selectors.R
```diff
@@ -65,18 +79,279 @@ selector__sub <- function(other) {
...
+      name = "and",
+      parameters = list(
+        self = self,
+        other = other
+      )
+    )
+  } else {
+    self$as_expr()$and(other)
+  }
+}
```
eitsupi, line 110: Do we need `is_column` to match the Python implementation?
https://github.com/pola-rs/polars/blob/d51b12c706983b423e8ad344a5636ae903fe3431/py-polars/polars/selectors.py#L386-L400

etiennebacher: I added this for `&` and `|` but there's one part in the `&` code that I'm not sure about: https://github.com/pola-rs/polars/issues/19740

R/selectors.R
```diff
@@ -101,3 +395,701 @@ cs__by_name <- function(..., .require_all = TRUE) {
...
+#' df$select(cs$contains("ba", "z"))
+#'
+#' # Select all columns except for those that contain the substring "ba":
+#' df$select(!cs$contains("ba"))
+cs__contains <- function(...) {
+  check_dots_unnamed()
+  input <- list2(...)
+  if (!all(vapply(input, is_character, FUN.VALUE = logical(1)))) {
+    abort("All elements of `...` must be of type character.")
+  }
```
eitsupi, line 446: I believe all inputs are checked when using vapply. Wouldn't it make more sense to use a for loop instead and generate an error the moment there is an unwanted input?

eitsupi (changes requested):

R/selectors.R
etiennebacher: I think that would make it more complex for very limited performance gains.

eitsupi: I meant: https://stackoverflow.com/questions/23142695/efficient-versions-of-any-all

etiennebacher: It makes almost no diff for me:
``` r
dat <- as.list(sample(letters, 10000, TRUE))
all_vapply <- function(x) {
  all(vapply(x, is.character, logical(1)))
}
loop <- function(x) {
  for (xi in x) {
    if (!is.character(xi)) {
      return(FALSE)
    }
  }
  TRUE
}
bench::mark(
  all_vapply = all_vapply(dat), 
  loop = loop(dat)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 all_vapply   2.26ms   2.49ms      389.   39.11KB     25.3
#> 2 loop       495.83µs 554.41µs     1742.    3.95MB     26.4

dat[[10001]] <- 1
bench::mark(
  all_vapply = all(vapply(dat, is.character, logical(1))), 
  loop = loop(dat)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 all_vapply   2.23ms   2.51ms      392.    39.1KB     26.3
#> 2 loop       525.54µs  573.3µs     1673.        0B     26.5
```

<sup>Created on 2024-11-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/selectors.R
etiennebacher: Fixed upstream so I made the same change here

R/selectors.R
eitsupi: If the loop is terminated in the middle of the process, it will be faster because no extra checks are performed.

``` r
dat <- as.list(sample(letters, 10000, TRUE))
all_vapply <- function(x) {
  all(vapply(x, is.character, logical(1)))
}
loop <- function(x) {
  for (xi in x) {
    if (!is.character(xi)) {
      return(FALSE)
    }
  }
  TRUE
}
bench::mark(
  all_vapply = all_vapply(dat),
  loop = loop(dat)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 all_vapply   4.53ms   4.84ms      194.   39.11KB     13.2
#> 2 loop         1.03ms   1.17ms      727.    3.95MB     11.0

dat[[5000]] <- 1
bench::mark(
  all_vapply = all(vapply(dat, is.character, logical(1))),
  loop = loop(dat)
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 all_vapply   4.51ms   4.92ms      185.    39.1KB     13.5
#> 2 loop       512.57µs 667.17µs     1260.        0B     10.6
```

<sup>Created on 2024-11-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher: That sort of confirms my point, this makes it less readable for a gain < 5ms when 10k patterns are passed, so I don't think this tradeoff is worth it.

R/selectors.R
```diff
@@ -982,7 +982,7 @@ cs__starts_with <- function(...) {
 
 #' Select all String (and, optionally, Categorical) string columns.
 #'
-#' @inheritParams rlang::check_dots_empty0
+#' @inheritParams rlang::args_dots_empty
```
etiennebacher, line 985: Why this change? Personally I prefer the "Dots which should be empty." from `check_dots_empty0()` (straight to the point) over "These dots are for future extensions and must be empty." from `args_dots_empty()` (explains why they should be empty, which isn't necessary IMO).

If we end up going with your change, it should be applied everywhere throughout the docs.

eitsupi: Because that is what `check_dots_empty` recommends.
https://rlang.r-lib.org/reference/check_dots_empty.html

> If we end up going with your change, it should be applied everywhere throughout the docs.

I already did that in 4d1b6993118f182e50ea28dd8556e114a2899ceb. (I had noticed this in the review phase of a past PR, but it wasn't a big deal so I just merged it in with the intention of fixing it myself later)

etiennebacher: Ok, thanks

R/selectors.R
```diff
@@ -519,27 +523,22 @@ cs__date <- function() {
...
+    abort("`time_zone` must be a character vector, `NULL`, or a list.")
   }
+  if (is.null(time_zone)) {
+    time_zone <- list(NULL)
+  }
+
+  datetime_dtypes <- time_unit |>
+    lapply(\(tu) {
+      time_zone |>
+        lapply(\(tz) pl$Datetime(tu, tz))
```
eitsupi, line 538: The error when an invalid time zone is entered here is fairly difficult to understand, so there may be future improvement potential.

eitsupi (approved):
Thanks, let's merge for now.

R/selectors.R
```diff
@@ -101,3 +387,702 @@ cs__by_name <- function(..., .require_all = TRUE) {
...
+#' # Select all datetime columns that have NO timezone:
+#' df$select(cs$datetime(time_zone = NULL))
+#'
+#' # Select all columns except for datetime columns:
+#' df$select(!cs$datetime())
+cs__datetime <- function(time_unit = c("ms", "us", "ns"), time_zone = list("*", NULL)) {
+  time_unit <- arg_match(time_unit, values = c("ms", "us", "ns"), multiple = TRUE)
+  if (!is_character(time_zone) && is_list(time_zone) && is.null(time_zone)) {
+    abort("`time_zone` must be a character vector, `NULL`, or a list.")
+  }
```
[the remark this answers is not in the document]
eitsupi: Indeed. Fix it so that it is caught by the test case.

R/selectors.R
etiennebacher: FYI this doesn't error in py-polars:

```
import polars as pl
import polars.selectors as cs
from datetime import datetime
from zoneinfo import ZoneInfo

tokyo_tz = ZoneInfo("Asia/Tokyo")

df = pl.DataFrame(
    {
        "tstamp_tokyo": [
            datetime(1999, 7, 21, 5, 20, 16, 987654, tzinfo=tokyo_tz),
            datetime(2000, 5, 16, 6, 21, 21, 123465, tzinfo=tokyo_tz),
        ]
    }
)

df.select(cs.datetime(time_zone = "foobar"))
```
Or did you mean for instance input like `time_zone = 1`?

etiennebacher: > Or did you mean for instance input like time_zone = 1?

Ah no that would be caught earlier

eitsupi: Oh, sorry for the lack of clarity. It means that if we pass a list as an argument, each element of the list must be a character of length 1.

```r
# error
cs$datetime(time_zone = list(c("UTC", "Asia/Tokyo")))
# ok
cs$datetime(time_zone = c("UTC", "Asia/Tokyo"))
```

PR #33, opened 2024-11-07
feat: add more range functions

etiennebacher:
Adds `time_range(s)` and `int_range(s)`


R/functions-range.R
```diff
@@ -219,3 +220,173 @@ pl__datetime_ranges <- function(
...
+#'     start = hms::parse_hms("14:00:00"),
+#'     interval = as.difftime("3:15:00")
+#'   )
+#' )
+pl__time_range <- function(
+    start = NULL,
+    end = NULL,
+    interval = "1h",
+    ...,
+    closed = "both") {
```
eitsupi, line 324:
```suggestion
    closed = c("both", "left", "none", "right")) {
```

R/functions-range.R
```diff
@@ -219,3 +220,173 @@ pl__datetime_ranges <- function(
...
+#'   start = hms::parse_hms(c("09:00:00", "10:00:00")),
+#'   end = hms::parse_hms(c("11:00:00", "11:00:00"))
+#' )
+#' df$with_columns(time_range = pl$time_ranges("start", "end"))
+pl__time_ranges <- function(
+    start = NULL,
+    end = NULL,
+    interval = "1h",
+    ...,
+    closed = "both") {
```
eitsupi, line 368:
```suggestion
    closed = c("both", "left", "none", "right")) {
```

R/functions-range.R
```diff
@@ -219,3 +220,173 @@ pl__datetime_ranges <- function(
...
+    for (unit in c("y", "mo", "w", "d")) {
+      if (grepl(unit, interval)) {
+        abort(paste0("invalid interval unit: found ", unit))
+      }
+    }
+    if (is.null(start)) {
+      start <- hms::parse_hms("00:00:00")
+    }
+    if (is.null(end)) {
+      end <- hms::parse_hms("23:59:59.999")
```
eitsupi, line 383: If you do this, it will not work if hms is not installed. Cast literal from Int32 to Time.

eitsupi (changes requested):

R/functions-range.R
etiennebacher: Forgot how to create time columns without hms, thanks

R/functions-range.R
```diff
@@ -219,3 +220,171 @@ pl__datetime_ranges <- function(
...
+    check_dots_empty0(...)
+    if (is.null(end)) {
+      end <- start
+      start <- 0
+    }
+    int_range(
+      as_polars_expr(start)$`_rexpr`,
+      as_polars_expr(end)$`_rexpr`,
+      step,
+      dtype$`_dt`
```
eitsupi, line 262: `dtype`'s type should be checked first. (`` dtype$`_dt` `` should be error)

eitsupi (changes requested):

R/functions-range.R
etiennebacher: Done

R/functions-range.R
```diff
@@ -219,3 +220,173 @@ pl__datetime_ranges <- function(
...
+#' @examplesIf requireNamespace("hms", quietly = TRUE)
+#' pl$select(
+#'   time = pl$time_range(
+#'     start = hms::parse_hms("14:00:00"),
+#'     interval = as.difftime("3:15:00")
+#'   )
+#' )
+pl__time_range <- function(
+    start = NULL,
+    end = NULL,
```
eitsupi, line 323: I feel there is no need to allow `NULL` here.
I think `start = pl$lit(0)$cast(pl$Time)` makes more sense.

etiennebacher: I disagree, I think:
* it would blur the fact that it's possible to pass something that is not an Expr as `start` and `end`
* for consistency we'd need to put `end = pl$lit(86399999999999)$cast(pl$Time)`, which is very unclear
* we'd need to check whether `start` is `NULL` anyway so it wouldn't save lines of code

(This also follows py-polars API btw)

eitsupi: I understand. Could you use the `%||%` operator in that case to match the general pattern in R?

etiennebacher: Done

eitsupi (approved):
Thanks!

PR #34, opened 2024-11-08
feat!: more methods of `struct` namespace

etiennebacher:


R/expr-struct.R
```diff
@@ -34,25 +65,103 @@ expr_struct_field <- function(...) {
...
+#' )
+expr_struct_rename_fields <- function(names) {
   wrap({
-    check_dots_unnamed()
-
-    dots <- list2(...) |>
-      unlist(recursive = FALSE)
-    check_character(dots, arg = "...", allow_na = FALSE)
-
-    self$`_rexpr`$struct_rename_fields(dots)
```
eitsupi, line 45: Why?

eitsupi (changes requested):

R/expr-struct.R
etiennebacher: It's not `*` in py-polars, it's a vector: https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.struct.rename_fields.html

And the input in Rust is `StringSexp` so I don't see why we'd use `...` here

eitsupi: I see. In that case, please use the check function to make sure there is no NA.

etiennebacher: Done

PR #37, opened 2024-11-09
feat: enable `list$concat()`

etiennebacher:


R/expr-list.R
```diff
@@ -154,10 +154,10 @@ expr_list_n_unique <- function() {
 #'   conc_to_lit_list = pl$col("a")$list$concat(pl$lit(list("hello", c("hello", "world"))))
 #' )
 expr_list_concat <- function(other) {
-  # TODO-REWRITE: py-polars internals are more complicated than that but need
-  # pl$concat_list() to be implemented before finishing this.
-  pl$concat_list(list(other)) |>
-    wrap()
+  wrap({
+    class(self) <- c("polars_expr", "polars_object")
```
etiennebacher, line 158: Feels hacky but otherwise `self` has a class `polars_namespace_expr` which breaks `as_polars_expr()` called downstream. Any idea to make this better?

eitsupi: > Any idea to make this better?

I don't understand what you are trying to say. I see no reason why we can't do the same thing here with Python Polars.

https://github.com/pola-rs/polars/blob/py-1.12.0/py-polars/polars/expr/list.py#L466-L503

```python
other_list.insert(0, wrap_expr(self._pyexpr))
```

etiennebacher: I think I just mixed up the things I tried, this works, thank you

eitsupi (approved):
Thanks!

PR #40, opened 2024-11-13
feat: finish implementation of `pl$concat()` except for the `align` strategy

etiennebacher:


R/functions-eager.R
```diff
@@ -1,25 +1,193 @@
...
+#' df2 <- pl$DataFrame(a = 2L, c = 4L)
+#' pl$concat(c(df1, df2), how = "diagonal")
+#'
+#' df_a1 <- pl$DataFrame(id = 1:2, x = 3:4)
+#' df_a2 <- pl$DataFrame(id = 2:3, y = 5:6)
+#' df_a3 <- pl$DataFrame(id = c(1L, 3L), z = 7:8)
+#' pl$concat(c(df_a1, df_a2, df_a3), how = "align")
+pl__concat <- function(
+    items,
+    ...,
```
etiennebacher, line 57: In current polars we pass items through dots but if we want to match py-polars we should only pass them via `items`

eitsupi: In my opinion, if something like list or dict is the first argument in Python, it should use dynamic dots in R.

- Consistency with functions such as `c()` should also be considered here.
- R lists are not identical to Python lists. There are some tricky points, such as the presence or absence of names, allowing multiple same names, etc. By using dynamic dots, these points can be easily checked and made consistent with many packages.

src/rust/src/conversion/mod.rs
```diff
@@ -96,6 +96,36 @@ impl TryFrom<ListSexp> for Wrap<Vec<DataFrame>> {
     }
 }
 
+impl TryFrom<ListSexp> for Wrap<Vec<LazyFrame>> {
+    type Error = savvy::Error;
+
+    fn try_from(list: ListSexp) -> Result<Self, savvy::Error> {
+        let dfs = list
```
eitsupi, line 103: They are not dataframes.

src/rust/src/conversion/mod.rs
```diff
@@ -96,6 +96,36 @@ impl TryFrom<ListSexp> for Wrap<Vec<DataFrame>> {
...
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(Wrap(dfs))
+    }
+}
+
+impl TryFrom<ListSexp> for Wrap<Vec<Series>> {
+    type Error = savvy::Error;
+
+    fn try_from(list: ListSexp) -> Result<Self, savvy::Error> {
+        let dfs = list
```
eitsupi, line 118: Ditto.

eitsupi (changes requested):

R/functions-eager.R
etiennebacher: I don't mind that, I updated to use dynamic dots

R/dataframe-frame.R
```diff
@@ -499,3 +499,41 @@ dataframe__sort <- function(
...
+#'
+#' df$n_chunks()
+#' df$n_chunks(strategy = "all")
+dataframe__n_chunks <- function(strategy = c("first", "all")) {
+  wrap({
+    strategy <- arg_match0(strategy, values = c("first", "all"))
+    if (strategy == "first") {
+      self$`_df`$n_chunks()
+    } else {
+      vapply(as.list(self, as_series = TRUE), \(x) x$n_chunks(), FUN.VALUE = integer(1))
```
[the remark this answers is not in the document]
etiennebacher: Two reasons:
1. it directly returns a vector, so no need to call `unlist()` at the end
2. it ensures the output of each iteration is as expected (integer here). If `n_chunks()` changes later, it would probably be caught elsewhere anyway, but I took the habit of being explicit about the output type when I can (it's also easier to read IMO).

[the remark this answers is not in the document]
eitsupi: Thank you. I am convinced.

R/functions-eager.R
```diff
@@ -1,25 +1,223 @@
...
+#' df2 <- pl$DataFrame(a = 2L, c = 4L)
+#' pl$concat(df1, df2, how = "diagonal")
+#'
+#' df_a1 <- pl$DataFrame(id = 1:2, x = 3:4)
+#' df_a2 <- pl$DataFrame(id = 2:3, y = 5:6)
+#' df_a3 <- pl$DataFrame(id = c(1L, 3L), z = 7:8)
+#' pl$concat(df_a1, df_a2, df_a3, how = "align")
+pl__concat <- function(
+    ...,
+    how = "vertical",
```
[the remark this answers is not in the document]
etiennebacher: It's long but also the values that `how` can take depend on the input type (which is why the `arg_match0()` call comes late and isn't at the top of the function), so I don't think we should list them here.

R/functions-eager.R
```diff
@@ -1,25 +1,223 @@
...
+        abort("Series only supports 'vertical' concat strategy.")
+      }
+    } else if (is_polars_expr(first)) {
+      return(
+        items |>
+          lapply(\(x) x$`_rexpr`) |>
+          concat_expr(rechunk = rechunk) |>
+          wrap()
+      )
+    } else {
```
eitsupi, line 212: This is indeed the same as Python Polars, but it looks rather odd not to check `how`......

etiennebacher: Yes, I think it's missing in py-polars. I'll add more checks tomorrow (I'm also unsure of what the accepted values are for Expr inputs).

R/functions-eager.R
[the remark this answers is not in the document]
eitsupi: I understand that point of view, but I think it might be worth putting somethings for readability (since `arg_match0` can check for different default values for actual arguments).
For example, the `.names_repare` argument of the `as_tibble` function accepts input that is different from the default value (e.x. `"unique_quiet"`).
https://github.com/eitsupi/neo-r-polars/blob/0c54776fa5ba04c76382b978dfad5d4eecb841b4/R/dataframe-s3-tibble.R#L28

R/functions-eager.R
etiennebacher: https://github.com/pola-rs/polars/issues/19813

R/functions-eager.R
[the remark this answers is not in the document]
etiennebacher: I don't understand your argument. My point is that if we list all values of `how`, someone might think `"horizontal"` is accepted for all input types while it doesn't work with Series and therefore it is misleading . The code you link to only runs for `polars_data_frame` so there is no situation in which one of the values won't work because of the input type. 

>  For example, the .names_repare argument of the as_tibble function accepts input that is different from the default value (e.x. "unique_quiet").

This is unclear to me, do you have an example?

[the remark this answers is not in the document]
eitsupi: > My point is that if we list all values of `how`, someone might think `"horizontal"` is accepted for all input types while it doesn't work with Series and therefore it is misleading

I understand that, but isn't that a problem because a friendly error message is displayed?

tests/testthat/test-functions-eager.R
eitsupi, line 1: I think we can write parameterized tests to reduce the number of lines.

eitsupi (commented):
I have made some fixes but it does not work perfectly because joins are not implemented.

R/functions-eager.R
etiennebacher: I have removed the code and tests related to Expr following the answer in the issue above.

tests/testthat/test-functions-eager.R
etiennebacher: I don't think this is worth it, it would make the test file shorter at the expense of readability and easiness to quickly identify the faulty tests. I'll try to add more parameterized tests in future PRs but for now I don't feel like rewriting this.

R/functions-eager.R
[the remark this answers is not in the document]
etiennebacher: Good point, I added the list of possible values in the function signature

etiennebacher:
Thanks, I don't plan on changing anything else so feel free to merge if you want

eitsupi (approved):
Thanks, I think the Expr concat could be revived, but for now it looks good to merge.

PR #42, opened 2024-11-18
feat: add `pl$scan_ipc()` and `pl$read_ipc()`

etiennebacher:
- [x] check examples
- ~read raw IPC~ Won't do it here, it's tricky on the Rust side because it goes to `get_mmap_bytes_reader()` which then uses pyo3 specific functions (I think)


R/io-ipc-functions.R
```diff
@@ -0,0 +1,142 @@
+#' Lazily read from an Arrow IPC (Feather v2) file or multiple files via glob
+#' patterns
+#'
+#' This allows the query optimizer to push down predicates and projections to
+#' the scan level, thereby potentially reducing memory overhead.
+#'
+#' @inheritParams rlang::args_dots_empty
+#' @inheritParams pl_scan_parquet
```
etiennebacher, line 8: This will be fixed when `$scan_parquet()` will be implemented

etiennebacher:
@eitsupi I'm always puzzled about the differences between the read and scan functions in py-polars API. For instance scan_ipc has two args for hive partitioning, but read_ipc has none. Do you know why?

(To me, this is one of the situations where it's ok to depart from py-polars API)

R/io-ipc-functions.R
```diff
@@ -0,0 +1,145 @@
...
+#' # If the path is a folder, Polars automatically tries to detect partitions
+#' # and includes them in the output
+#' pl$scan_ipc(temp_dir)$collect()
+#'
+#' # We can also impose a schema to the partition
+#' pl$scan_ipc(temp_dir, hive_schema = list(cyl = pl$String, gear = pl$Int32))$collect()
+pl__scan_ipc <- function(
+    source,
+    ...,
+    columns = NULL,
```
eitsupi, line 53: Is this needed?

eitsupi (commented):
Thanks for working on this!

> Do you know why?

I think it is because of the different origins of those functions.
However, since it is planned to unify them (pola-rs/polars#13040), I do not see a problem with implementing only `scan_*` to simplify implementation.

R/io-ipc-functions.R
etiennebacher: This appears in read_ipc but not scan_ipc. IMO it's not super useful since it's a simple select() call at the end. No problem to remove if you want.

tests/testthat/test-io-ipc.R
```diff
@@ -0,0 +1,200 @@
...
+
+  # cannot get a subset of partitioning columns
+  expect_error(
+    pl$scan_ipc(temp_dir, hive_schema = list(cyl = pl$String))$collect(),
+    "field not found: path contains column not present in the given Hive schema"
+  )
+})
+
+test_that("try_parse_hive_dates works", {
+  skip_if_not_installed("arrow")
```
eitsupi, line 156: Please add `arrow` to the Suggests field.

tests/testthat/test-io-ipc.R
```diff
@@ -0,0 +1,200 @@
...
+  skip_if_not_installed("withr")
```
eitsupi, line 157: Because `testthat` is based on `withr`, I don't think this line is useful.

R/io-ipc-functions.R
```diff
@@ -54,19 +44,21 @@ pl__scan_ipc <- function(
     row_index_name = NULL,
     row_index_offset = 0L,
     storage_options = NULL,
-    memory_map = TRUE,
```
etiennebacher: Why remove this argument? I don't see any mentions of deprecation in the docs

eitsupi: This does not actually do anything.
That is probably a bug in the Python documentation, which has already been removed from the Rust side.
See pola-rs/polars#18622

etiennebacher: Alright, thanks

PR #44, opened 2024-11-21
feat: add `pl$scan_parquet()` and `pl$read_parquet()`

etiennebacher:


src/rust/src/lazyframe/general.rs
```diff
@@ -290,4 +290,102 @@ impl PlRLazyFrame {
...
+            cloud_options: None,
+            use_statistics,
+            schema: schema.map(|x| Arc::new(x)),
+            hive_options,
+            glob,
+            include_file_paths: include_file_paths.map(|x| x.into()),
+            allow_missing_columns,
+        };
+
+        let first_path = source.first().unwrap().clone().into();
```
eitsupi, line 366: Do we need to check the length on the R side or raise an error without using `unwrap()` here?

eitsupi (approved):
Looks good, thanks!
(I'm sorry, but I don't have time today, so please wait a little while to update #30)

One minor thing:

src/rust/src/lazyframe/general.rs
etiennebacher: No problem for the time.

Indeed, we also panick in the IPC functions if the user passes `source = character(0)`. I'll fix that.

PR #47, opened 2024-11-30
docs: fix math notation in `emw_` functions

etiennebacher:
Addresses https://github.com/eitsupi/neo-r-polars/pull/19#discussion_r1864119287

Now:

image

### DIFF
--- a/R/expr-expr.R
+++ b/R/expr-expr.R
@@ -3322,8 +3322,8 @@ expr__rolling_var_by <- function(
 #' account for imbalance in relative weightings:
 #' * when `TRUE` (default), the EW function is calculated using weights
 #'  \eqn{w_i = (1 - \alpha)^i};
-#' * when `FALSE`, the EW function is calculated recursively by \deqn{y_0 &= x_0
-#'  \\y_t &= (1 - \alpha)y_{t - 1} + \alpha x_t}
+#' * when `FALSE`, the EW function is calculated recursively by \deqn{y_0 = x_0
+#'  ; y_t = (1 - \alpha)y_{t - 1} + \alpha x_t}
 #' @param bias If `FALSE` (default), apply a correction to make the estimate
 #' statistically unbiased.
 #' @param ignore_nulls Ignore missing values when calculating weights.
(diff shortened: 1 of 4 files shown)


eitsupi:
Thanks!

eitsupi (commented):
It still seems to be broken.

eitsupi:
@etiennebacher What do you think about removing the latex notation altogether? I do not see the importance of this.

eitsupi:
Delete all of these for now.

eitsupi:
This is causing the R-universe build to keep failing. This document is not a priority for me and will be removed until I am sure it can be built.

image

eitsupi:
> Can you open an issue with this and more details about what doesn't work? Then I can take a look at this

Ultimately the problem is that there is no CI set up for PRs.
Since such a complex problem cannot be immediately verified at hand, it can be removed and done later rather than forcing it at this stage.

etiennebacher:
Where do you see errors related to the math notation? In https://github.com/r-universe/eitsupi/actions/runs/12096400448/job/33730857076 I only see warnings about undocumented args or missing crosslinks

eitsupi:
> Where do you see errors related to the math notation?

https://github.com/r-universe/eitsupi/actions/runs/12096400448/job/33730657339#step:4:1655

eitsupi:
Perhaps it is here that it is broken.

image

etiennebacher:
Can you revert https://github.com/eitsupi/neo-r-polars/commit/c5bb6416681caed57a28ab1f86881cab956bef9e#diff-51f523bfdc0ef8541ff83534df89bf98d71331ce0714a756b76ee487b90da381L3439 to only use simple comments + a `TODO` instead of removing the entire block? It's very annoying to copy and modify from py-polars docs and I plan on fixing it later.

eitsupi:
ab15898131fe7a32fc164fb08fd94daadd2db05f

etiennebacher:
Thanks, I'll take a look later this weekend

eitsupi:
I appreciate you working on this, but I don't think I have enough knowledge to do a review and therefore can't merge.
(As a result, it was completely premature to rush into merging this pull request.)

PR #49, opened 2024-12-14
feat: add missing functions and docs for LazyFrame

etiennebacher:
Same changes as in #31, just without the tests


src/rust/src/conversion/mod.rs
```diff
@@ -4,10 +4,10 @@ use crate::prelude::*;
 use crate::{PlRDataFrame, PlRDataType, PlRExpr, PlRLazyFrame, PlRSeries, RPolarsErr};
 use polars::prelude::cloud::CloudOptions;
 use polars::series::ops::NullBehavior;
-use savvy::{ListSexp, NumericScalar, NumericSexp, NumericTypedSexp, StringSexp, TypedSexp};
+use savvy::{ListSexp, NumericScalar, NumericSexp, NumericTypedSexp, Sexp, StringSexp, TypedSexp};
 use search_sorted::SearchSortedSide;
 pub mod base_date;
-mod chunked_array;
+pub mod chunked_array;
```
eitsupi, line 10: Why?

src/rust/src/conversion/mod.rs
```diff
@@ -33,6 +33,25 @@ impl<T> From<T> for Wrap<T> {
...
+            TypedSexp::String(x) => {
+                let val = x.to_vec();
+                AnyValue::StringOwned((*val.first().unwrap()).into())
+            }
+            TypedSexp::Null(_) => AnyValue::Null,
+            _ => return Err("Cannot cast to AnyValue".to_string()),
+        };
+        Ok(Wrap(out))
+    }
+}
```
eitsupi, line 53: What is this?
Do not add conversions that do not involve `as_polars_series()`. (i.e., this should have to be removed)

eitsupi (changes requested):

eitsupi:
Thanks, but uust to be clear, don't expect to see this huge change merged immediately as it is very difficult to review.

R/lazyframe-frame.R
```diff
@@ -358,12 +755,1924 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#' @inherit as_polars_lf return
+#' @examples
+#' lf <- pl$LazyFrame(a = 1:6, b = 7:12)
+#' lf$limit()$collect()
+#' lf$limit(2)$collect()
+lazyframe__limit <- function(n = 5) {
+  wrap({
+    self$head(n)
+  })
+}
```
eitsupi, line 785:
```suggestion
#' @rdname lazyframe__head
lazyframe__limit <- lazyframe__head
```

R/lazyframe-frame.R
```diff
@@ -358,12 +755,1924 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+    simplify_expression = TRUE,
+    slice_pushdown = TRUE,
+    comm_subplan_elim = TRUE,
+    comm_subexpr_elim = TRUE,
+    cluster_with_columns = TRUE,
+    streaming = FALSE,
+    no_optimization = FALSE,
+    collect_in_background = FALSE,
+    show_plot = FALSE,
+    truncate_nodes = 0) {
```
eitsupi, line 1447: Should not allow positional arguments.

R/lazyframe-frame.R
```diff
@@ -358,12 +755,1924 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'   attr(data, "created_on") <- "2024-01-29"
+#'   data
+#' }
+#' df1 <- as_polars_lf(iris)
+#' df2 <- give_attr(df1)
+#'
+#' # now, the original LazyFrame doesn't get this attribute
+#' attributes(df1)
+lazyframe__clone <- function() {
+  self$`_ldf`$clone()
```
eitsupi, line 1570:
```suggestion
  self$`_ldf`$clone() |>
    wrap()
```

R/lazyframe-frame.R
```diff
@@ -358,12 +755,1924 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+  })
+}
+
+#' Add an external context to the computation graph
+#'
+#' This allows expressions to also access columns from DataFrames or LazyFrames
+#' that are not part of this one.
+#'
+#' @param other Data/LazyFrame to have access to. This can be a list of DataFrames
+#' and LazyFrames.
```
eitsupi, line 1611: Is this really true?

R/lazyframe-frame.R
```diff
@@ -358,12 +755,1924 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'   a = c(NA, 2, 3, 4),
+#'   b = c(0.5, NA, 2.5, 13),
+#'   c = c(TRUE, TRUE, FALSE, NA)
+#' )
+#'
+#' df$clear()
+#'
+#' df$clear(n = 5)
+lazyframe__clear <- function(n = 0) {
+  pl$DataFrame(schema = self$schema)$clear(n)$lazy()
```
eitsupi, line 1958: Is this correct???

eitsupi (changes requested):

src/rust/src/conversion/mod.rs
etiennebacher: This is the way it is in py-polars:

https://github.com/pola-rs/polars/blob/91ad299b17c48422131c584f63acedc758581100/crates/polars-python/src/lazyframe/general.rs#L992C27-L992C31

There is no other places in our codebase where an input is `Wrap<AnyValue<'_>>` so far and it's very rare in the py-polars coddebase:

https://github.com/search?q=repo%3Apola-rs%2Fpolars+Wrap%3CAnyValue%3C%27_%3E%3E&type=code

R/lazyframe-frame.R
etiennebacher: `with_context()` was actually deprecated in 1.0, I'll remove it

R/lazyframe-frame.R
etiennebacher: It needs a small adjustment since we don't have `schema` but `.schema_overrides` in `pl$DataFrame()` but that's the idea, yes:

https://github.com/pola-rs/polars/blob/91ad299b17c48422131c584f63acedc758581100/py-polars/polars/lazyframe/frame.py#L3238

(Note that this doesn't work because `$clear()` is not implemented for DataFrame yet)

etiennebacher: [`a1b5f16` (#49)](https://github.com/eitsupi/neo-r-polars/pull/49/commits/a1b5f162f0e2ee3a50fb683f5e1b4a78991f8519)

src/rust/src/conversion/mod.rs
eitsupi: Check carefully: does Python Polars do type mapping in more than one place?
The way the type mapping is done here is fundamentally different from that in polars-python. So this must be removed.

src/rust/src/lazyframe/serde.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[savvy]
+fn deserialize_lf(json: &str) -> Result<PlRLazyFrame> {
+    let lp = serde_json::from_str::<DslPlan>(json).map_err(|_| {
+        let msg = "could not deserialize input into a LazyFrame";
+        RPolarsErr::Other(msg.to_string())
+    })?;
+    let out = LazyFrame::from(lp);
+    Ok(<PlRLazyFrame>::from(out))
+}
```
eitsupi, line 12: This is not a clone of Python Polars' ones, is it?
The function should be the same as Python Polars.

Of course, it must be an associated function.

eitsupi (changes requested):

src/rust/src/conversion/mod.rs
eitsupi: To add a note that may help your understanding, in Python Polars, Python objects are converted to Polars literals via AnyValue. But that is not the case here.

etiennebacher: I don't understand what you mean. In `join_asof()`'s options, `tolerance` must be of type `AnyValue` so we will have to convert the input to an `AnyValue` in Rust anyway, no?
```rust
pub struct AsOfOptions {
    pub strategy: AsofStrategy,
    /// A tolerance in the same unit as the asof column
    pub tolerance: Option<AnyValue<'static>>,
    ...
}
```
Usually I can figure this out by looking at functions that you already implemented in neo-r-polars, but as I said this is the only place where `AnyValue` is used as input. I don't see how to change this, so can you point me more precisely to the change to make?

src/rust/src/lazyframe/serde.rs
etiennebacher: No I think I had copied this from r-polars actually, so it should be modified/removed.

Deserialization is a bit tricky in terms of API and I think that's why we have deviated with this `deserialize_lf()` in r-polars. In py-polars they have:

```
pl.DataFrame.deserialize(...)
pl.LazyFrame.deserialize(...)
```

and I'm not sure neo-r-polars support something like this since `pl$DataFrame` just returns the body of the function. I wouldn't mind removing this serde file for now, what do you think?

src/rust/src/conversion/mod.rs
etiennebacher: Great, thank you 👍

src/rust/src/lazyframe/serde.rs
eitsupi: Since `pl$DataFrame` is a function, something like `pl$DataFrame$foo` cannot be realized.

However, this is a different thing: whether the function on the R side is something like `deserialize_lf`, an associated function can be used on the Rust side (extendr binds the associated functiona to the R side in a strange way, but with savvy, it is not a problem).

src/rust/src/conversion/mod.rs
eitsupi: However, there is a problem, it seems that if I pass something like `list(NULL)`, there is a panic.
This is not seen when Python Polars installed from PyPI. I have not had time to dig into the cause.

etiennebacher: https://github.com/pola-rs/polars/issues/20321

eitsupi: Thanks for taking a look at this. I did my tests on Python with DataFrame, so perhaps this hints that turning off the collect method optimization will not result in panic.

src/rust/src/lazyframe/serde.rs
etiennebacher: I'm not sure I understand and I think this might be tricky to implement. I have removed this function for now: it's not essential and I don't want it to block the rest of the PR. I added a todo and we can address this later in a more focused PR.

etiennebacher:
@eitsupi just checking in to know if you were able to review parts of this PR already. I know this one is big, I could maybe split it in 2 or 3 if it helps.

I'll also try to fix the wasm build errors. Do you know if there's a way to mock this build locally or if it can only be done in CI?

eitsupi:
Sorry for the late review. I have not read through most of the individual methods.
I'm pretty stuck on the API, especially `StatisticsOptions`, and I think it would be more natural for R to have each element as a separate argument rather than specifying it as a list.

> I'll also try to fix the wasm build errors. Do you know if there's a way to mock this build locally or if it can only be done in CI?

This would be relatively easy to run locally as it builds on Docker, but I have not tried it.

https://github.com/r-wasm/actions/tree/main/build-rwasm

etiennebacher:
>  I'm pretty stuck on the API, especially StatisticsOptions, and I think it would be more natural for R to have each element as a separate argument rather than specifying it as a list.

I'm not convinced about this for several reasons. First, it increases the number of arguments to browse in the docs, and maybe more statistics will be added in the future so it would be annoying to have 5-6 arguments only about statistics. Second, we use lists in other params as well (e.g. `storage_options` in `sink_parquet()`) so I don't see why we would diverge here.

Other packages like `readr` have functions like [`locale()`](https://readr.tidyverse.org/reference/locale.html) which are not very useful on their own but can be passed to `read_csv()` for instance. I think it would be a good compromise: it allows to have proper named arguments while not adding extra args to `sink_parquet()`. What do you think about this?

etiennebacher:
@eitsupi Thoughts on the previous comment?

eitsupi:
Sorry, I forgot to comment.

Your opinion makes sense, and I think it's a good idea to pass the elements to Rust that decompose the list on the R side.

etiennebacher:
I've added a `parquet_statistics()` function which is essentially a `list()` but allows us to ensure there are no misnamed statistics. I'd like to keep the TRUE/FALSE shortcut to include or exclude all stats at once.

We can discuss this further but if so I'd rather put a TODO on this and address this in a subsequent PR so we can focus on the other methods in this PR. Once this one is merged I can move on with the tests and the DataFrame methods (which I will do in smaller pieces to avoid massive PRs like this one).

eitsupi (changes requested):
Could you please remove all IO-related functions for several reasons?

- It is too complex to review in a short time.
- Break WASM builds.
- On the Rust side, I believe the file should be split (see pola-rs/polars#19937)

etiennebacher:
@eitsupi all IO functions should be removed now

R/lazyframe-frame.R
```diff
@@ -358,12 +769,1397 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'
+#' lf$join(other_lf, on = "ham", how = "left", coalesce = TRUE)$collect()
+#'
+#' lf$join(other_lf, on = "ham", how = "semi")$collect()
+#'
+#' lf$join(other_lf, on = "ham", how = "anti")$collect()
+lazyframe__join <- function(
+    other,
+    on = NULL,
+    how = "inner",
```
eitsupi, line 1156: I may have asked this somewhere before, but is there any reason not to list them all here?

R/lazyframe-frame.R
```diff
@@ -358,12 +769,1397 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+    how <- arg_match0(
+      how,
+      values = c("inner", "full", "left", "right", "semi", "anti", "cross")
+    )
+    validate <- arg_match0(validate, values = c("m:m", "1:m", "m:1", "1:1"))
+    uses_on <- !is.null(on)
+    uses_left_on <- !is.null(left_on)
+    uses_right_on <- !is.null(right_on)
+    uses_lr_on <- uses_left_on | uses_right_on
+    if (uses_on && uses_lr_on) {
```
eitsupi, line 1178: I think type checks are needed.

R/lazyframe-frame.R
```diff
@@ -358,12 +769,1397 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'   sum_a = pl$col("a")$sum(),
+#'   min_a = pl$col("a")$min(),
+#'   max_a = pl$col("a")$max()
+#' )$collect()
+lazyframe__rolling <- function(
+    index_column,
+    ...,
+    period,
+    offset = NULL,
+    closed = "right",
```
eitsupi, line 1515: I think all values to be listed.

R/lazyframe-frame.R
```diff
@@ -358,12 +769,1397 @@ lazyframe__slice <- function(offset, length = NULL) {
...
+#'   closed = "right"
+#' )$agg(A_agg_list = pl$col("A"))$collect()
+lazyframe__group_by_dynamic <- function(
+    index_column,
+    ...,
+    every,
+    period = NULL,
+    offset = NULL,
+    include_boundaries = FALSE,
+    closed = "left",
```
eitsupi, line 1678: Ditto.

R/lazyframe-frame.R
etiennebacher: I don't understand what you mean. This is just to check exclusivity of `on` and `left_on` / `right_on`. Why should we check types here instead of letting this to the Rust part? 

Even if we did check, I'm not even sure what should be accepted or not. For instance this works:

```r
lf$join(other_lf, left_on = 1, right_on = 2)$collect()

# shape: (0, 5)
# ┌─────┬─────┬─────┬───────┬───────────┐
# │ foo ┆ bar ┆ ham ┆ apple ┆ ham_right │
# │ --- ┆ --- ┆ --- ┆ ---   ┆ ---       │
# │ i32 ┆ f64 ┆ str ┆ str   ┆ str       │
# ╞═════╪═════╪═════╪═══════╪═══════════╡
# └─────┴─────┴─────┴───────┴───────────┘
```
and this fails:
```r
lf$join(other_lf, left_on = TRUE, right_on = 2)$collect()

# Error:
# ! Evaluation failed in `$collect()`.
# Caused by error:
# ! datatypes of join keys don't match - `_POLARS_0`: bool on left does not match `_POLARS_1`: f64 on right

# Resolved plan until failure:
# 
# 	---> FAILED HERE RESOLVING 'join' <---
#  WITH_COLUMNS:
#  [2.0.alias("_POLARS_1")] 
#   DF ["apple", "ham"]; PROJECT */2 COLUMNS
```

Same behavior in py-polars.

R/lazyframe-frame.R
etiennebacher: Thanks, done in [`3f0af7f` (#49)](https://github.com/eitsupi/neo-r-polars/pull/49/commits/3f0af7f40ce5bc40b8e62c44ca69d739888cf573) (also for the two other comments).

> I may have asked this somewhere before, but is there any reason not to list them all here?

I think we decided to go with this before but it's just not in my habits. I don't really see the usecase but we already discussed this so no need to do it more ;)

R/lazyframe-frame.R
eitsupi: Sorry, you are right. I think I missed a few lines above.

eitsupi (approved):
Sorry it took so long to confirm. I appreciate your work and patience.
Let's merge this for now.

etiennebacher:
Thanks for the useful review, I'll start making PRs for DataFrame methods (only 2-3 per PR).

PR #52, opened 2025-01-04
fix: Rename args of `str$pad_start()`, `str$pad_end()`, and `str$strip_chars_*()`

etiennebacher:
Noticed this thanks to the dashboard, this was changed a long time ago in py-polars.


src/rust/src/expr/string.rs
```diff
@@ -68,16 +68,16 @@ impl PlRExpr {
...
     }
 
-    fn str_pad_start(&self, width: NumericScalar, fillchar: &str) -> Result<Self> {
-        let width = <Wrap<usize>>::try_from(width)?.0;
-        let fillchar = <Wrap<char>>::try_from(fillchar)?.0;
-        Ok(self.inner.clone().str().pad_start(width, fillchar).into())
+    fn str_pad_start(&self, length: NumericScalar, fill_char: &str) -> Result<Self> {
+        let length = <Wrap<usize>>::try_from(length)?.0;
+        let fill_char = <Wrap<char>>::try_from(fill_char)?.0;
+        Ok(self.inner.clone().str().pad_start(length, fill_char).into())
```
etiennebacher, line 80: Argument names for the `strip_*` functions were not changed in rust-polars: https://github.com/pola-rs/polars/blob/7fddd840e9d37d1b62024f395c0ea43fc6938375/crates/polars-python/src/expr/string.rs#L76

eitsupi (approved):
Thanks!

PR #54, opened 2025-01-04
fix: Rename args of `str$starts_with()`, `str$ends_with()`, `str$zfill()`

etiennebacher:


src/rust/src/expr/string.rs
```diff
@@ -59,13 +59,8 @@ impl PlRExpr {
...
 
-    fn str_zfill(&self, alignment: &PlRExpr) -> Result<Self> {
-        Ok(self
-            .inner
-            .clone()
-            .str()
-            .zfill(alignment.inner.clone())
-            .into())
+    fn str_zfill(&self, length: &PlRExpr) -> Result<Self> {
+        Ok(self.inner.clone().str().zfill(length.inner.clone()).into())
```
etiennebacher, line 63: Arg names in Rust didn't change for starts_with() and ends_with()

R/expr-string.R
```diff
@@ -407,16 +410,10 @@ expr_str_strip_chars_end <- function(matches = NULL) {
...
-#' # cast to String and ljust alignment = 5, and view as R char vector
-#' some_floats_expr$cast(pl$String)$str$zfill(5)$to_r()
-#'
-#' # cast to int and the to utf8 and then ljust alignment = 5, and view as R
-#' # char vector
-#' some_floats_expr$cast(pl$Int64)$cast(pl$String)$str$zfill(5)$to_r()
-expr_str_zfill <- function(alignment) {
-  self$`_rexpr`$str_zfill(as_polars_expr(alignment)$`_rexpr`) |>
+#' sdf <- pl$DataFrame(a = c(-1, 123, 999999, NA))
+#' df$with_columns(zfill = pl$col("a")$cast(pl$String)$str$zfill(4))
```
etiennebacher, line 414: Old example that outdated in terms of functions and unclear. The new example is the one in py-polars

R/expr-string.R
```diff
@@ -407,16 +410,10 @@ expr_str_strip_chars_end <- function(matches = NULL) {
...
-#' # cast to String and ljust alignment = 5, and view as R char vector
-#' some_floats_expr$cast(pl$String)$str$zfill(5)$to_r()
-#'
-#' # cast to int and the to utf8 and then ljust alignment = 5, and view as R
-#' # char vector
-#' some_floats_expr$cast(pl$Int64)$cast(pl$String)$str$zfill(5)$to_r()
-expr_str_zfill <- function(alignment) {
-  self$`_rexpr`$str_zfill(as_polars_expr(alignment)$`_rexpr`) |>
+#' df <- pl$DataFrame(a = c(-1L, 123L, 999999L, NA))
+#' df$with_columns(zfill = pl$col("a")$cast(pl$String)$str$zfill(4))
```
etiennebacher, line 414: Old example was outdated in terms of functions and unclear. The new example is the one in py-polars

src/rust/src/expr/string.rs
eitsupi: In that case, the error message is unclear if an argument type mismatch occurs on the Rust side (which may not happen here).
Could you update the Rust side argument names to match the R side?

etiennebacher: Done

eitsupi (approved):
Thanks!

PR #55, opened 2025-01-04
feat: Add `str$to_decimal()`

etiennebacher:


R/expr-string.R
```diff
@@ -420,6 +420,27 @@ expr_str_zfill <- function(alignment) {
...
+#'
+#' @examples
+#' df <- pl$DataFrame(
+#'   numbers = c(
+#'     "40.12", "3420.13", "120134.19", "3212.98",
+#'     "12.90", "143.09", "143.9"
+#'   )
+#' )
+#' df$with_columns(numbers_decimal = pl$col("numbers")$str$to_decimal())
+expr_str_to_decimal <- function(inference_length = 100) {
```
eitsupi, line 440: At first glance it is unclear what kind of role the number in the first argument plays, so for me it is surprising that this argument is not a named-only argument.

This is fine for now, but we probably need to update Python Polars.

eitsupi (approved):
Thanks!

R/expr-string.R
eitsupi: > we probably need to update Python Polars.

I've opened an issue pola-rs/polars#20564

PR #59, opened 2025-01-07
feat: support `pl$arg_where()` and `<expr>$arg_true()`

etiennebacher:


R/expr-expr.R
```diff
@@ -1510,16 +1510,16 @@ expr__arg_unique <- function() {
...
-# pl$arg_where(self$`_rexpr`)
-# }
+#' Return indices where expression is true
+#'
+#' @inherit as_polars_expr return
+#' @examples
+#' df <- pl$DataFrame(a = c(1, 1, 2, 1))
+#' df$select((pl$col("a") == 1)$arg_true())
+expr__arg_true <- function() {
+  pl$arg_where(self) |>
```
eitsupi, line 1520:
```suggestion
  arg_where(self) |> 
```

eitsupi (changes requested):
Thanks. I've compared with Python Polars.

R/expr-expr.R
etiennebacher: Yes, good catch

PR #61, opened 2025-01-07
feat: add `pl$Field()` 

etiennebacher:
This is a bit tricky because `pl$Field()` needs to be a `DataType` so that it can be parsed in `parse_into_list_of_datatypes()` in `pl$Struct()`, but it's not part of the `DataType` enum in rust-polars.

This works, but maybe you have a cleaner way in mind.

### DIFF
--- a/R/000-wrappers.R
+++ b/R/000-wrappers.R
@@ -2697,13 +2697,6 @@ class(`PlRDataType`) <- c("PlRDataType__bundle", "savvy_neopolars__sealed")
   }
 }
 
-`PlRExpr_str_json_decode` <- function(self) {
-  function(`dtype`, `infer_schema_len`) {
-    `dtype` <- .savvy_extract_ptr(`dtype`, "PlRDataType")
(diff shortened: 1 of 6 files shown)


etiennebacher:
> Is `pl$Field` worth existing? In my opinion it is not worth existing, and I don't think it exists in Python Polars either.

It exists but is not documented. It appears in the examples of `st.json_decode()`. I can open an issue upstream and then split or close the PR depending on their answer.

eitsupi (changes requested):
Thank you. Some further thoughts:

- See the definition of the Field class of Python Polars. It does not inherit DataType. In other words, it is completely different from what you write here.
- Why is this class necessary? I think it is better not to have it because it only complicates things (in fact, it has such a role in Python Polars that no one would notice if it were not documented)

etiennebacher:
It's true it doesn't seem essential, especially since we can use e.g. `pl$Struct(a = pl$Float64, b = pl$Boolean)` in the test. I'm fine with letting this aside.

PR #63, opened 2025-01-08
docs: `rolling_*_by()` support integers

etiennebacher:
Also add more tests for those


R/expr-expr.R
```diff
@@ -2862,8 +2862,9 @@ expr__rolling <- function(
 #' * …
 #' * `(t_n - window_size, t_n]`
 #'
-#' @param by This column must be of dtype Datetime or Date. Accepts expression
-#' input. Strings are parsed as column names.
+#' @param by Should be DateTime, Date, UInt64, UInt32, Int64, or Int32 data 
+#' type (note that the integer ones require using `"i"` in `window size`). 
+#' Accepts expression input. Strings are parsed as column names.
```
eitsupi, line 2867: It would be great if we could explain well that "something is converted to expression of data type DateTime, Date, UInt64, UInt32, Int64 or Int32 by `as_polars_expr()`."

eitsupi (changes requested):

R/expr-expr.R
etiennebacher: How about 

> Should be DateTime, Date, UInt64, UInt32, Int64, or Int32 data type after conversion by `as_polars_expr()`. Note that integer types require using...

?

I didnt' check the rest of the docs but this wording might be applicable in other places as well.

eitsupi: Thanks, looks good.

> I didnt' check the rest of the docs but this wording might be applicable in other places as well.

Definitely. Improving documentation is an afterthought.......

R/expr-expr.R
```diff
@@ -2862,8 +2862,10 @@ expr__rolling <- function(
 #' * …
 #' * `(t_n - window_size, t_n]`
 #'
-#' @param by This column must be of dtype Datetime or Date. Accepts expression
-#' input. Strings are parsed as column names.
+#' @param by Should be DateTime, Date, UInt64, UInt32, Int64, or Int32 data 
+#' type after conversion by [`as_polars_expr()`][as_polars_expr]. Note that the
```
eitsupi, line 2866:
```suggestion
#' type after conversion by [as_polars_expr]. Note that the
```

eitsupi (changes requested):

R/expr-expr.R
etiennebacher: Done, I just kept the parenthesis otherwise the link doesn't work anymore

eitsupi: Oh sorry, my mistake. Thanks.

eitsupi (approved):
Thanks!

PR #64, opened 2025-01-08
feat: Add argument `as_dot` in `$meta$tree_format()`

etiennebacher:
I think we talked before about whether we should include the options to print / save the graph using external packages. I don't remember the decision, but no problem to add / remove functionalities here.

### DIFF
--- a/R/expr-meta.R
+++ b/R/expr-meta.R
@@ -192,6 +192,7 @@ expr_meta_root_names <- function() {
 
 #' Format the expression as a tree
 #'
+#' @inheritParams rlang::args_dots_empty
 #' @param as_dot If `TRUE`, show the dot syntax that can be used in other
 #' packages, such as `DiagrammeR`.
 #'
@@ -208,9 +209,11 @@ expr_meta_root_names <- function() {
 #' graph <- my_expr$meta$tree_format(as_dot = TRUE)
 #' DiagrammeR::grViz(graph)
 #' }
-expr_meta_tree_format <- function(as_dot = FALSE) {
-  self$`_rexpr`$compute_tree_format(display_as_dot = as_dot) |>
-    wrap()
+expr_meta_tree_format <- function(..., as_dot = FALSE) {
+  wrap({
+    check_dots_empty0(...)
+    self$`_rexpr`$compute_tree_format(display_as_dot = as_dot)
+  })
 }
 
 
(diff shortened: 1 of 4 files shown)


eitsupi:
> I think we talked before about whether we should include the options to print / save the graph using external packages. I don't remember the decision

Maybe https://github.com/pola-rs/r-polars/pull/928#discussion_r1528630691?

Related upstream issue: pola-rs/polars#20631

etiennebacher:
In the `r-polars` issue we decided to go with `$to_dot()` so that we don't have these extra `Suggests`, but this function is [not exported in py-polars](https://docs.pola.rs/api/python/dev/search.html?q=to_dot).

At the same time, if we keep `$show_graph()` then it's weird to not print a graph but only the dot syntax. In the end I think adding those 3 optional deps only matters for CI time.

eitsupi:
My concern is not the dependency but the API.
In other words, I don't want complex combinations of arguments to result in completely different types of output.

In my opinion, we should separate functions rather than controlling the type of output with arguments.
IIUC, this is known as the [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy).

etiennebacher:
> In my opinion, we should separate functions rather than controlling the type of output with arguments.

I'd agree if we were building our own API but here we're following an existing one that faces the same challenges. I think we could instead talk about this upstream, especially with the `mermaid` PR that adds an additional layer of complexity to the function in Python. Having `.to_dot()`, `.to_mermaid()` and `.show_graph()` without the `raw_output` args would make sense IMO.

eitsupi:
Fair point.
As you say, given the purpose here, I think `show_graph` should be for display only (Like `polars.show_versions()`).

I was really worried about this API and had left the following comment in the past (i.e. I had postponed adding this method because I wasn't sure about the better API).

https://github.com/eitsupi/neo-r-polars/blob/ed926df831e130e01f82f1ad61fdc68e71afb22e/R/expr-meta.R#L189

etiennebacher:
Since there doesn't seem to be much activity for this function upstream, I'm proposing a custom argument `as_dot` in `meta$tree_format()` and I dropped `meta$show_graph()`. 

I think this is a good compromise: 
* it allows printing the tree as usual but gives the opportunity to get the dot syntax to use it in other packages, such as DiagrammeR
* it keeps the type stability of the output (always return a string)
* we don't add extra dependencies in Suggests

eitsupi (approved):
Thanks!

PR #67, opened 2025-01-09
feat: add `$flags` for `DataFrame` and `Series`

etiennebacher:


R/series-series.R
```diff
@@ -46,6 +46,21 @@ wrap.PlRSeries <- function(x, ...) {
...
+    "flags",
+    function() {
+      out <- list(
+        SORTED_ASC = self$`_s`$is_sorted_ascending_flag(),
+        SORTED_DESC = self$`_s`$is_sorted_descending_flag()
+      )
+      # TODO: how can we have self$dtype == pl$List() ?
+      # if (self$dtype$eq(pl$List())) {
+      #   out[["FAST_EXPLODE"]] <- self$`_s`$can_fast_explode_flag()
+      # }
```
etiennebacher, line 59: This is the annoying part

R/dataframe-frame.R
```diff
@@ -98,6 +112,21 @@ wrap.PlRDataFrame <- function(x, ...) {
...
+    function() {
+      out <- lapply(
+        # Polars is 0-indexed
+        seq_along(self$columns) - 1,
+        function(x) {
+          self$to_series(x)$flags
+        }
+      )
+      names(out) <- self$columns
+      out
```
eitsupi, line 126: How about something like `lapply(self$get_columns(), \(x) x$flags)`

etiennebacher: Sure, I forgot about `$get_columns()`

R/series-series.R
```diff
@@ -46,6 +46,20 @@ wrap.PlRSeries <- function(x, ...) {
...
+    function() {
+      out <- list(
+        SORTED_ASC = self$`_s`$is_sorted_ascending_flag(),
+        SORTED_DESC = self$`_s`$is_sorted_descending_flag()
+      )
+      if (inherits(self$dtype, "polars_dtype_list")) {
+        out[["FAST_EXPLODE"]] <- self$`_s`$can_fast_explode_flag()
+      }
+      out
+    },
```
eitsupi, line 60: Shouldn't this return a logical vector instead of a list?

eitsupi (changes requested):

R/dataframe-frame.R
```diff
@@ -98,6 +112,15 @@ wrap.PlRDataFrame <- function(x, ...) {
   makeActiveBinding("shape", function() self$`_df`$shape(), self)
   makeActiveBinding("height", function() self$`_df`$height(), self)
   makeActiveBinding("width", function() self$`_df`$width(), self)
+  makeActiveBinding(
+    "flags",
+    function() {
+      out <- lapply(self$get_columns(), \(x) x$flags)
+      names(out) <- self$columns
+      out
```
eitsupi, line 120: I don't think `names(out) <- self$columns` is necessary since the list is originally named.

R/series-series.R
etiennebacher: We could, yes. No strong preference on this

eitsupi (changes requested):

R/series-series.R
eitsupi: Thanks, I understand. There is no need to use list here, so please use logical.

etiennebacher: Done

eitsupi (approved):
Looks good, thanks!

PR #50, opened 2024-12-30
fix: support WASM/webR builds

eitsupi:

PR #69, opened 2025-01-12
chore: update rust deps with `cargo update` and downgrade cc from 1.2.8 to 1.2.7

eitsupi:
Ref: https://github.com/eitsupi/neo-r-polars/pull/50#issuecomment-2585308954, rust-lang/cc-rs#1362

I've just run:

```sh
$ pushd src/rust
$ cargo update
$ cargo update cc --precise 1.2.7
```


eitsupi:
cc 1.2.8 was confirmed as the cause of the build error.

PR #73, opened 2025-01-23
build(ci/taskfile): Add CI and Task for formatting R and Rust code

### LINKED ISSUE #72: Add CI and task for formatting
Now that [`air` 1.0.0](https://github.com/posit-dev/air/releases/tag/1.0.0) is released, should we add a workflow for this? I have one here: https://github.com/etiennebacher/tidypolars/blob/main/.github/workflows/format.yml

We can also add a `task format` that would run it and `cargo fmt`

etiennebacher:
Fixes #72

Putting this in draft until a few issues and features are complete in `air`, specifically:

* ~ignore files generated by `savvy`: https://github.com/posit-dev/air/issues/128#issuecomment-2610021341~ ==> Done
* ~formatting for calls to `tribble()`: https://github.com/posit-dev/air/issues/113. Currently those are completely expanded vertically. Alternatively, add suppression comments around those (https://github.com/posit-dev/air/issues/52) depending on which of those features lands first in `air`~ ==> Done
```r
tibble::tribble(
  ~.test_name, ~object,
  "Int8", pl$Int8,
  "Int16", pl$Int16,
  "Int32", pl$Int32,
  "Int64", pl$Int64,
  "UInt8", pl$UInt8,
)
```
becomes:
```r
tibble::tribble(
  ~.test_name,
  ~object,
  "Int8",
  pl$Int8,
  "Int16",
  pl$Int16,
  "Int32",
  pl$Int32,
  "Int64",
  pl$Int64,
  "UInt8",
  pl$UInt8,
)
```

---

When those are fixed, I'll make a separate PR where I actually run the formatters so that this PR is not drowned in formatting changes.


Taskfile.yml
```diff
@@ -158,3 +158,20 @@ tasks:
     desc: Accept all test snapshots. (Shortcut to accept snapshots after running tests)
     cmds:
       - Rscript -e 'testthat::snapshot_accept()'
+
+  install-air:
+    desc: Install Air to format R code.
+    cmds:
+      - curl -LsSf https://github.com/posit-dev/air/releases/latest/download/air-installer.sh | sh
```
etiennebacher, line 165: This is downloaded every time `task fmt` runs, which is bad. @eitsupi do you know why this is executed while other setup tasks are skipped? Is it because this is a curl command and there's no way to know if it's outdated?

eitsupi:
I saw this post.
<https://www.tidyverse.org/blog/2025/02/air/>

Can we use the GitHub Action used by the shiny team?
<https://github.com/rstudio/shiny-workflows/tree/7114cb3f0e1c3d05af27c6668b980362f1fb0864/format-r-code>

etiennebacher:
Sure I'll edit that

etiennebacher:
I answered a bit too quickly. I'm not a big fan of automatically committing fixes from `air`. It regularly happens that I push several commits only separated by a few seconds, and having automatic commits would make conflicts because I'd need to pull locally and push again. I'd rather have a "check-only" workflow.

Also, the workflow in this PR runs both cargo fmt and air. Should we separate those in two workflows?

eitsupi:
Oh, sorry. I didn't take a good look at what's in their workflow. I don't like auto-commit either. I have no problem with linter-only execution.

Taskfile.yml
```diff
@@ -158,3 +158,14 @@ tasks:
...
       - Rscript -e 'testthat::snapshot_accept()'
+
+  fmt:
+    desc: Format all R and Rust code.
+    sources:
+      - "{{.R_SOURCE}}"
+      - "{{.RUST_SOURCE}}"
+      - tests/**/*
+    cmds:
+      - air -V | grep -q "0.4.0" || exit 1
```
etiennebacher, line 169: Only way I found to ensure `task fmt` errors if `air` is not at the right version

air.toml
```diff
@@ -0,0 +1,7 @@
+[format]
+line-width = 80
+indent-width = 2
+indent-style = "space"
+line-ending = "auto"
+persistent-line-breaks = true
+exclude = ["R/000-wrappers.R"]
```
eitsupi, line 7: Doesn't the latest Air automatically skip formatting of auto-generated files?

eitsupi:
Could the CI error indicate a bug in Air's parsing?

```log
ERROR Failed to format tests/testthat/test-as_polars_series.R: Failed to parse due to syntax errors.
ERROR Failed to format tests/testthat/test-datatypes-classes.R: Failed to parse due to syntax errors.
ERROR Failed to format tests/testthat/test-expr-string.R: Failed to parse due to syntax errors.
```

air.toml
etiennebacher: Oh yes indeed

etiennebacher:
> Could the CI error indicate a bug in Air's parsing?

No the CI errors simply because there are files that would be reformatted (but maybe I misunderstood your question?)

eitsupi:
I am not sure why CI is getting an error. Are you saying that you have edited some R files but have not yet run the auto-format?

eitsupi:
(I'm going to bed and will look at it tomorrow.)

etiennebacher:
Yes I set up the CI ans task but haven't run formatting yet. I thought about doing in another PR so its easier to review this one.

eitsupi:
I have tried running the Air CLI locally and I still think these errors are bugs in the parser.

```sh
$ air format .
ERROR Failed to format tests/testthat/test-as_polars_series.R: Failed to parse due to syntax errors.
ERROR Failed to format tests/testthat/test-expr-string.R: Failed to parse due to syntax errors.
ERROR Failed to format tests/testthat/test-datatypes-classes.R: Failed to parse due to syntax errors.
```

I think it would be better to open another PR for the Air setup, so please hold on a moment.

etiennebacher:
> I still think these errors are bugs in the parser.

Yes they probably are but I think they are tracked in `tree-sitter-r`. I don't see those errors as blockers for this PR.

eitsupi:
Just to be clear, I am not saying it will be a PR blocker, just that I suspect they need to be excluded for now.

etiennebacher:
I'm ok with that, do you want to update #135 to exclude them?

eitsupi:
Yeah, but just as I wrote in posit-dev/air#255, these all seem to be caused by raw Strings, so it might be better to update the R file.

Taskfile.yml
eitsupi: Perhaps it is the design of Taskfile to set `preconditions` here, but in any case, it is not common to fix the version of VS Code extensions, so for now, it would be better to assume automatic updates of the latest version?

https://taskfile.dev/usage/#using-programmatic-checks-to-cancel-the-execution-of-a-task-and-its-dependencies

etiennebacher:
There are no longer parsing failures

tests/testthat/test-as_polars_series.R
```diff
@@ -79,7 +79,7 @@ test_that("as_polars_series.default throws an error", {
 test_that("as_polars_series.polars_expr throws an error", {
   expect_error(
     as_polars_series(pl$lit(1)),
-    r"(You can evaluating the expression with `pl\$select\(\))"
+    "You can evaluating the expression with `pl\\$select\\(\\)"
```
eitsupi, line 82: From my digging, the problem here seems to be a mismatch in the number of backticks, and I think we can keep the Raw strings.

etiennebacher: Done

Taskfile.yml
```diff
@@ -158,3 +159,21 @@ tasks:
...
+      - exclude: R/000-wrappers.R
+      - exclude: R/import-standalone-*.R
+    generates:
+      - "{{.R_SOURCE}}"
+      - "{{.R_TESTS}}"
+      - exclude: R/000-wrappers.R
+      - exclude: R/import-standalone-*.R
+    cmds:
+      - air format .
+      - cargo fmt --all --manifest-path src/rust/Cargo.toml
```
etiennebacher, line 179: I'd rather have a single `task fmt` for both R and Rust

eitsupi: Since deps are generally executed in parallel, it is better to split the task than to execute them sequentially in cmds.

etiennebacher: Ok but can we still have a `task fmt` that calls `task fmt-r` and `task fmt-rust`?

eitsupi: I've cherry picked 66eee85, and will add tasks to format Rust files and format all files

etiennebacher: Both of those are extremely fast anyway so running sequentially doesn't change much

eitsupi: Yeah, just another problem is the task dependencies: since the R files and Rust files are each dependencies of a completely different subsequent tasks, it makes more sense to format them separately than to format them at the same time.

eitsupi: Done in 4d9aa0ef771877262ddbf5a773c5989d1f861fb1.

etiennebacher: Thanks, I think it can be merged now

eitsupi:
I would like to make some more tweaks to the workflow files before merging.
(Please wait a few more hours as I am going out to eat.)

etiennebacher:
Sure, I won't be available before tomorrow anyway

eitsupi (approved):
Although a few changes in the R files have been included, they are minor and let's merge them in as they are.

Thanks for working on this!

PR #74, opened 2025-01-27
feat!: bump polars to rs-0.46.0

eitsupi:


tests/testthat/test-expr-string.R
```diff
@@ -979,16 +979,17 @@ test_that("$str$tail works", {
 
 test_that("$str$extract_many works", {
   df <- pl$DataFrame(values = c("discontent", "dollar $"))
-  patterns <- pl$lit(c("winter", "disco", "ONTE", "discontent", "$"))
+  # FIXME: The literal must be a single list[str] after <https://github.com/pola-rs/polars/pull/20781>
+  patterns <- pl$lit(list(c("winter", "disco", "ONTE", "discontent", "$")))
```
eitsupi, line 983: @etiennebacher  As I wrote https://github.com/pola-rs/polars/pull/20781#discussion_r1930609364, your PR seems break this test.
Is that intended?
How should this function work?

etiennebacher: Thank you, good catch, this is indeed an issue in my `polars` PR. I'll open a new one upstream to fix this.

eitsupi: Thanks, I see pola-rs/polars#20953

PR #78, opened 2025-02-14
feat: Add `$with_columns()` and `$with_columns_seq()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -297,28 +297,90 @@ dataframe__select <- function(...) {
...
 #' @inherit as_polars_df return
 #' @examples
-#' as_polars_df(iris)$with_columns(
-#'   abs_SL = pl$col("Sepal.Length")$abs(),
-#'   add_2_SL = pl$col("Sepal.Length") + 2
+#' # Pass an expression to add it as a new column.
+#' df <- pl$DataFrame(
+#'   a = 1:4,
+#'   b = c(0.5, 4, 10, 13),
+#'   c = c(TRUE, TRUE, FALSE, TRUE),
```
etiennebacher, line 304: Just using the same examples as in lazy `with_columns()`

eitsupi (commented):
Thanks, I suspect now that we have the LazyFrame and DataFrame methods (`with_columns()` and `with_columns_seq()`), it would be a good time to add some tests.
Any thoughts on adding tests?

etiennebacher:
You mean adding tests for those two in this PR as well or in a separate one?

eitsupi:
I think it would be better if the relevant tests were added in this PR.
Ideally, it would be tested in CI (of course, right now there is no tests for PRs because it is far from ideal, but it will be tested on the R-universe once merged).

Of course if you want to improve the overall test you were previously working on first, you can add the test afterwards.

etiennebacher:
I'll add some tests here

tests/testthat/helper-expections.R
etiennebacher, line 1: Improvements that were made here: https://github.com/eitsupi/neo-r-polars/pull/31/files#diff-62a08fcd787a5161df91277268cfbae6e2515c1cde86e968161a362fd2982164

Not strictly needed for this PR, mostly useful when there are several inputs (e.g. for joins) but I thought it would be ok to add it here since I start working on tests

tests/testthat/test-lazyframe-frame.R
etiennebacher, line 1: Are we still ok to store all "frame" tests here and have the `expect_query_equal()` test for both lazy and eager?

Some methods might be eager-only or lazy-only so it's still good to keep a "test-dataframe-frame" I think.

eitsupi (approved):
Thanks!

eitsupi:
(I didn't think too much about it, but as you know, the upstream polars repo require PR titles to start with a uppercase letter, so perhaps it would be better to follow that rule?)

eitsupi:
I think it would be better for the future to keep the commit history of this repository (i.e. currently the `next` branch of r-polars) (that's why I remove the PR number every time I squash-merge PRs in this repository).

I do not think merging into the current r-polars `main` branch is worthwhile as there is too little in the codebase shared, and I think a strategy of renaming branches and switching default branches would be better (see examples such as docker compose v1 and v2).

etiennebacher:
>  I think a strategy of renaming branches and switching default branches would be better

Makes sense, I forgot about this approach, I think it's a good idea.

PR #81, opened 2025-02-14
feat: Add `unique()` for DataFrame

etiennebacher:


R/lazyframe-frame.R
```diff
@@ -1021,7 +1021,7 @@ lazyframe__drop_nulls <- function(subset = NULL) {
   })
 }
 
-#' Drop duplicate rows from this DataFrame
```
etiennebacher, line 1024: Should have been "LazyFrame", but I removed it so that we can inherit docs in eager

eitsupi (approved):
I'm sorry, but I've been out for two days and can only merge from my mobile, so please resolve the conflict.

etiennebacher:
Ah ok every time I fix conflicts it needs another approval

PR #82, opened 2025-02-14
feat!: Rewrite `<lazyfeame/dataframe>$drop_nulls()`

etiennebacher:


R/functions-aggregation-horizontal.R
```diff
@@ -8,7 +8,7 @@
 #'
 #' @examples
 #' df <- pl$DataFrame(
-#'   a = c(1, 8, 3)
+#'   a = c(1, 8, 3),
```
etiennebacher, line 11: Small drive-by fixes

R/lazyframe-frame.R
```diff
@@ -1000,7 +1000,7 @@ lazyframe__reverse <- function() {
 #' @examples
 #' lf <- pl$LazyFrame(
 #'   foo = 1:3,
-#'   bar = c(6, NA, 8),
+#'   bar = c(6L, NA, 8L),
 #'   ham = c("a", "b", NA)
```
etiennebacher, line 1004: Otherwise the `cs$integer()` call below is useless

R/lazyframe-frame.R
```diff
@@ -1015,7 +1015,11 @@ lazyframe__reverse <- function() {
 lazyframe__drop_nulls <- function(subset = NULL) {
   wrap({
     if (!is.null(subset)) {
-      subset <- parse_into_list_of_expressions(!!!subset)
+      if (is_polars_selector(subset)) {
+        subset <- parse_into_list_of_expressions(!!!list(subset))
+      } else {
+        subset <- parse_into_list_of_expressions(!!!subset)
+      }
```
etiennebacher, line 1022: Fix when a polars selector is passed. Without this we have an error "can't splice an environment"

eitsupi (changes requested):
I'm sorry I haven't reviewed it enough, but it seems to me that dynamic-dots should be used here.

etiennebacher:
Actually both `drop_nans()` and `drop_nulls()` only have a `subset` argument:

https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.drop_nans.html
https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.drop_nulls.html

and on the top of my head and I don't have examples where py-polars has `subset` but we use dynamic dots.

eitsupi:
I think it's simpler to use dynamic-dots for functions that accept variable-length names, and I think some of the existing ones do.

etiennebacher:
I'm fine with that, I added dots

eitsupi (approved):
Thanks for quick updates!

PR #83, opened 2025-02-14
feat: Add `drop_nans()` for LazyFrame and DataFrame

etiennebacher:


src/rust/src/lazyframe/general.rs
```diff
@@ -626,6 +626,12 @@ impl PlRLazyFrame {
         Ok(ldf.drop_nulls(subset).into())
     }
 
+    fn drop_nans(&self, subset: Option<ListSexp>) -> Result<Self> {
+        let ldf = self.ldf.clone();
+        let subset = subset.map(|e| <Wrap<Vec<Expr>>>::from(e).0);
+        Ok(ldf.drop_nans(subset).into())
+    }
+
```
etiennebacher, line 634: Meaningful change, the rest is just `cargo fmt` doing its stuff

PR #92, opened 2025-02-14
feat!: Rewrite `<lazyframe/dataframe>$fill_null()`

etiennebacher:
https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.fill_null.html#polars.LazyFrame.fill_null


eitsupi (approved):
Good catch, thanks!

eitsupi:
I think we need a generic function something like `nanoarrow::infer_nanoarrow_schema()` to detect mapped polars dtype.

etiennebacher:
I don't follow, this doesn't seem to map an R type to a polars dtype?

``` r
nanoarrow::infer_nanoarrow_schema(1L)
#> <nanoarrow_schema int32>
#>  $ format    : chr "i"
#>  $ name      : chr ""
#>  $ metadata  : list()
#>  $ flags     : int 2
#>  $ children  : list()
#>  $ dictionary: NULL
```

eitsupi:
@etiennebacher This is what I was trying to say #118.

PR #108, opened 2025-02-15
feat: Add `pl$repeat_()`

etiennebacher:
`repeat` is a keyword, hence the name `repeat_()`.

Bunch of irrelevant changes due to autoformatting in `expr-expr.R`, you can ignore this file.


eitsupi:
Thanks!
I wonder if we can define `pl$rep()` as an alias.

etiennebacher:
I dont really see the point, I think it would be confusing to have "rep", "repeat_", and "repeat_by" in the code suggestions

eitsupi (approved):
Ok, let's merge this for now.

R/functions-repeat.R
```diff
@@ -0,0 +1,33 @@
...
+#' # Specify an output dtype
+#' pl$select(pl$repeat_(3, n = 3, dtype = pl$Int8))
+pl__repeat_ <- function(value, n, ..., dtype = NULL) {
+  wrap({
+    check_polars_dtype(dtype, allow_null = TRUE)
+    if (is_integerish(n)) {
+      if (n < 0) {
+        abort("`n` must be greater than or equal to 0.")
+      }
+      n <- pl$lit(n)$cast(pl$Int64)
```
eitsupi, line 28: The `dtype` option is not working, is this intentional?

etiennebacher: I don't know what you mean, `dtype` is used on the output, not on `n`, there's an example in the tests.

https://github.com/pola-rs/polars/blob/5e51218b9a11703d6872ef5ff1929e4e32108fc6/py-polars/polars/functions/repeat.py#L139

eitsupi: My bad, thanks.

PR #109, opened 2025-02-16
feat: Add `unpivot()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1201,3 +1201,33 @@ dataframe__join_where <- function(
...
+dataframe__unpivot <- function(
+    on = NULL,
+    ...,
+    index = NULL,
+    variable_name = NULL,
+    value_name = NULL) {
+  wrap({
+    check_dots_empty0(...)
+    # TODO: add selectors handling when py-polars' _expand_selectors() has moved
+    # to Rust
```
etiennebacher, line 1225: https://github.com/pola-rs/polars/blob/4eb4fb7527d26d3ecf9b80cc4ad92809064f8c35/py-polars/polars/selectors.py#L113

Also not currently handled in r-polars

PR #111, opened 2025-02-17
feat: Add `to_dummies()` for DataFrame

etiennebacher:
Only exists for DataFrame


R/dataframe-frame.R
```diff
@@ -1230,3 +1230,40 @@ dataframe__unpivot <- function(
...
+#'   bar = c(3, 4),
+#'   ham = c("a", "b")
+#' )
+#' df$to_dummies()
+#'
+#' df$to_dummies(drop_first = TRUE)
+# df$to_dummies(cs$integer(), separator=":")
+# df$to_dummies(cs$integer(), drop_first = TRUE, separator = ":")
+dataframe__to_dummies <- function(
+    columns = NULL,
```
eitsupi, line 1255: As with some other places, I suspect that this is better to use dynamic-dots.

etiennebacher: Good catch, I was in auto mode. Fixed

eitsupi (approved):

R/dataframe-frame.R
```diff
@@ -1230,3 +1230,50 @@ dataframe__unpivot <- function(
...
+    ...,
+    separator = "_",
+    drop_first = FALSE) {
+  # TODO: add selectors handling when py-polars' _expand_selectors() has moved
+  # to Rust (and update examples above)
+  wrap({
+    check_dots_unnamed()
+
+    dots <- list2(...)
+    check_list_of_string(dots, arg = "...")
```
eitsupi, line 1265: Ref:

https://github.com/eitsupi/neo-r-polars/blob/dce16ac13ec178db1c149fe2db01ab5a6a55717f/NEWS.md?plain=1#L170-L186

etiennebacher: I don't like this check because the error message is:

>  `...` must be a list of single strings, not a  list.

which isn't clear. The previous message was clearer IMO:

> `...` only accepts column names.

eitsupi: Yeah, the error messages could use improvement.

My point was that R's character vector is difficult to treat as column names without NA checks, etc.

PR #117, opened 2025-02-18
docs: Fix or comment out examples

etiennebacher:
I think failures [in R-universe actions](https://github.com/r-universe/eitsupi/actions/runs/13381257175) are at least partly due to failing examples.

I'd like the version available on R-universe to be updated so that the dashboard properly reflects what is already done and what's left to do, hence this PR.

I'll fix formatting changes.


R/expr-meta.R
```diff
@@ -243,10 +243,10 @@ expr_meta_is_column_selection <- function(..., allow_aliasing = FALSE) {
 #' e <- pl$col("foo")
 #' e$meta$is_column()
 #'
-#' e <- pl.col("foo") * pl.col("bar")
+#' e <- pl$col("foo") * pl$col("bar")
 #' e$meta$is_column()
 #'
-#' e <- pl.col(r"^col.*\d+$")
+#' e <- pl$col("^col\\.*\\d+$")
```
etiennebacher, line 249: I just suspect people are more familiar with escaping in regex than with raw strings

eitsupi (approved):
Thanks for working on this!

PR #118, opened 2025-02-18
feat: New generic function `infer_polars_dtype()`

eitsupi:
Add a function to get the dtype of the Series constructed as a result of `as_polars_series()` inexpensively.


R/infer_polars_dtype.R
```diff
@@ -0,0 +1,128 @@
...
+#' This function is similar to [nanoarrow::infer_nanoarrow_schema()].
+#'
+#' S3 objects based on atomic vectors or classes built on the vctrs package will work accurately
+#' if the S3 method of the [as_polars_series] function is defined.
+#' @inheritParams as_polars_series
+#' @param x An R object.
+#' @return A [polars DataType][DataType]
+#' @seealso
+#' - [as_polars_series()]
+#' @exmaples
```
etiennebacher, line 16: typo

R/infer_polars_dtype.R
```diff
@@ -0,0 +1,128 @@
...
+#' @rdname infer_polars_dtype
+#' @export
+infer_polars_dtype.POSIXlt <- infer_polars_dtype_default_impl
+
+#' @rdname infer_polars_dtype
+#' @export
+infer_polars_dtype.NULL <- function(x, ...) {
+  if (missing(x)) {
+    abort("The `x` argument of `infer_polars_dtype()` can't be missing")
+  }
```
etiennebacher, line 66: Are there cases where we end up in this method if `x` is missing?

src/rust/src/datatypes.rs
```diff
@@ -145,15 +145,15 @@ impl PlRDataType {
...
+        precision: Option<NumericScalar>,
+    ) -> Result<Self> {
         let precision = precision
             .map(<Wrap<usize>>::try_from)
             .transpose()?
             .map(|p| p.0);
-        let scale = scale
-            .map(<Wrap<usize>>::try_from)
-            .transpose()?
-            .map(|s| s.0);
```
eitsupi, line 156: I was surprised that VS Code's Rust extension does not support "formatting only selections"......

R/infer_polars_dtype.R
```diff
@@ -0,0 +1,128 @@
...
+  pl$Null
+}
+
+#' @rdname infer_polars_dtype
+#' @export
+infer_polars_dtype.list <- function(x, ..., strict = FALSE) {
```
etiennebacher, line 72: Should this parameter be named `.strict`?

tests/testthat/test-infer_polars_dtype.R
```diff
@@ -0,0 +1,56 @@
...
+    skip_if_not_installed("blob")
+    skip_if_not_installed("bit64")
+    skip_if_not_installed("vctrs")
+    skip_if_not_installed("clock")
+
+    withr::with_timezone(
+      "UTC",
+      list2(
+        as_polars_series(1),
+        pl$DataFrame(a = 1L, b = "foo"),
```
etiennebacher, line 14: Include an example for LazyFrame too?

eitsupi: Yes, ideally. But at the moment I can't add it here because `as_polars_series(<LazyFrame>)` fails.
Add a TODO comment.

R/infer_polars_dtype.R
eitsupi: I believe that is not necessary.

R/infer_polars_dtype.R
eitsupi: I think I learned this pattern from `tibble:::as_tibble.NULL`.
Try running `tibble::as_tibble()`.

R/infer_polars_dtype.R
eitsupi: Good catch.

eitsupi (commented):
Thanks for your review.
I will update.

R/infer_polars_dtype.R
etiennebacher: Right, I missed the fact those were not dynamic dots

R/infer_polars_dtype.R
etiennebacher: I see. I think it would also make sense to put it before the call to `UseMethod()` but I'm fine with that

R/infer_polars_dtype.R
```diff
@@ -0,0 +1,131 @@
...
+#' @inheritParams as_polars_series
+#' @param x An R object.
+#' @return A [polars DataType][DataType]
+#' @seealso
+#' - [as_polars_series()]
+#' @examples
+#' infer_polars_dtype(1:10)
+#'
+#' # The type inference is fast even if the object is huge
+#' infer_polars_dtype(1:100000000)
```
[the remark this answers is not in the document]
eitsupi: My comment may not even be appropriate. I wanted to express a comparison of the followings:

```r
x <- 1:100000000

tictoc::tic()
as_polars_series(x)$dtype
tictoc::toc()
#> Int32
#> 11.017 sec elapsed

tictoc::tic()
infer_polars_dtype(x)
tictoc::toc()
#> Int32
#> 0.124 sec elapsed
```

PR #119, opened 2025-02-18
feat: Add `transpose()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1316,3 +1316,52 @@ dataframe__partition_by <- function(..., maintain_order = TRUE, include_key = TR
...
+#'
+#' # Use an existing column as the new column names
+#' df <- pl$DataFrame(id = c("i", "j", "k"), a = c(1, 2, 3), b = c(4, 5, 6))
+#' df$transpose(column_names = "id")
+#' df$transpose(include_header = TRUE, header_name = "new_id", column_names = "id")
+dataframe__transpose <- function(
+    ...,
+    include_header = FALSE,
+    header_name = "column",
+    column_names = NULL) {
```
etiennebacher, line 1354: py-polars allows `column_names` to be an iterator but it wasn't clear to me how to adapt that in R (see in their examples):
https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.transpose.html#polars.DataFrame.transpose

src/rust/src/dataframe/general.rs
```diff
@@ -52,6 +53,28 @@ impl PlRDataFrame {
...
+            ))
+        } else {
+            None
+        };
+        let out = self
+            .df
+            .transpose(keep_names_as, new_col_names)
+            .map_err(RPolarsErr::from)?;
+        Ok(out.into())
+    }
```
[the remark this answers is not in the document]
eitsupi: I asked LLM to produce something similar to the Python Generator function example[^1] in R.

[^1]: <https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/py-polars/polars/dataframe/frame.py#L4692-L4707>

``` r
name_generator <- local({
  base_name <- "my_column_"
  count <- 0
  function() {
    name <- paste0(base_name, count)
    count <<- count + 1
    name
  }
})

name_generator()
#> [1] "my_column_0"
name_generator()
#> [1] "my_column_1"
name_generator()
#> [1] "my_column_2"
```

<sup>Created on 2025-02-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Not sure if this is useful here......

R/dataframe-frame.R
eitsupi: I asked LLM to produce something similar to the Python Generator function example[^1] in R.

[^1]: <https://github.com/pola-rs/polars/blob/4e286d8c83b0fcb56f0a7ea06d2eb731f179e01e/py-polars/polars/dataframe/frame.py#L4692-L4707>

``` r
name_generator <- local({
  base_name <- "my_column_"
  count <- 0
  function() {
    name <- paste0(base_name, count)
    count <<- count + 1
    name
  }
})

name_generator()
#> [1] "my_column_0"
name_generator()
#> [1] "my_column_1"
name_generator()
#> [1] "my_column_2"
```

<sup>Created on 2025-02-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Not sure if this is useful here......

etiennebacher: I didn't dig into this in detail but I think the problem is that the counter isn't reset when you call the function several times:

``` r
name_generator <- local({
  base_name <- "my_column_"
  count <- 0
  function() {
    name <- paste0(base_name, count)
    count <<- count + 1
    name
  }
})

foo <- function(x = name_generator()) {
  print(x)
}

foo()
#> [1] "my_column_0"

foo()
#> [1] "my_column_1"
```

src/rust/Cargo.toml
```diff
@@ -12,6 +12,7 @@ panic = "unwind"
 
 [dependencies]
 ciborium = "0.2"
+either = "1.13.0"
```
eitsupi, line 15:
```suggestion
either = "1.11"
```

See https://github.com/pola-rs/polars/blob/fb28c07c612f89398436e597cf842b99037bd58d/Cargo.toml#L39

eitsupi (changes requested):

R/dataframe-frame.R
```diff
@@ -1698,3 +1698,52 @@ dataframe__rolling <- function(
...
+    check_dots_empty0(...)
+    keep_names_as <- if (isTRUE(include_header)) {
+      header_name
+    } else {
+      NULL
+    }
+    if (is.null(column_names)) {
+      column_names <- character()
+    }
+    self$`_df`$transpose(keep_names_as = keep_names_as, column_names = column_names)
```
eitsupi, line 1747:
```suggestion
    self$`_df`$transpose(keep_names_as = keep_names_as, column_names = column_names %||% character())
```

eitsupi (changes requested):

R/dataframe-frame.R
eitsupi: Anyway, when done in R, it is natural to use a function that takes a vector and returns a vector.

eitsupi (approved):

R/dataframe-frame.R
```diff
@@ -1698,3 +1698,60 @@ dataframe__rolling <- function(
...
+    check_dots_empty0(...)
+    keep_names_as <- if (isTRUE(include_header)) {
+      header_name
+    } else {
+      NULL
+    }
+    if (is.function(column_names)) {
+      column_names <- column_names(seq_len(nrow(self)) - 1)
+    }
+    self$`_df`$transpose(keep_names_as = keep_names_as, column_names = column_names %||% character())
```
eitsupi, line 1755: If you give the `header_name` something other than a string or NULL, the name displayed in the error here would be `keep_name_as`.
So the argument name of the Rust side should be `header_name`

R/dataframe-frame.R
```diff
@@ -1741,6 +1749,9 @@ dataframe__transpose <- function(
     } else {
       NULL
     }
+    if (is.function(column_names)) {
```
eitsupi, line 1752: Could you use `is_function` instead?

R/dataframe-frame.R
etiennebacher: Or should we just add `check_character(header_name, allow_null = TRUE)`?

etiennebacher: Actually that wouldn't take into account length errors. I'll just rename the arg in rut

eitsupi: Yes, using `check_string` here is fine.

etiennebacher: Ah I forgot we had `check_string()` on top of `check_character()`

R/dataframe-frame.R
```diff
@@ -1698,3 +1698,61 @@ dataframe__rolling <- function(
...
+    check_dots_empty0(...)
+    keep_names_as <- if (isTRUE(include_header)) {
+      check_string(header_name, allow_null = TRUE)
+      header_name
+    } else {
+      NULL
+    }
+    if (is_function(column_names)) {
+      column_names <- column_names(seq_len(nrow(self)) - 1)
+    }
```
eitsupi, line 1755: How about checking with `is_character(n = nrow(self))` that the return value is char vec of length `nrow(self)`?

etiennebacher: Done

eitsupi (approved):
Thanks!

PR #121, opened 2025-02-18
feat: Add `sink_parquet()` for LazyFrame

etiennebacher:
Currently some `expect_silent()` fail because `sink_parquet()` warns about the transition to the new streaming engine


R/lazyframe-frame.R
```diff
@@ -2313,4 +2313,134 @@ lazyframe__join_asof <- function(
...
+      statistics <- parquet_statistics(
+        min = FALSE,
+        max = FALSE,
+        distinct_count = FALSE,
+        null_count = FALSE
+      )
+    }
+    if (!inherits(statistics, "polars_parquet_statistics")) {
+      abort("`statistics` must be TRUE, FALSE, or a call to `parquet_statistics()`.")
+    }
```
etiennebacher, line 2422: I think this is a good compromise: it allows the user to pass a logical or a more detailed list of statistics to write, while ensuring the user doesn't pass unknown stats. Handling this on the R side is also easier than doing it in Rust.

tests/testthat/test-lazyframe-frame.R
etiennebacher, line 1: We also have test files for IO, e.g. `testthat/io-parquet-functions.R` so those might be stored there.

R/lazyframe-frame.R
eitsupi: Agreed.
However, since `polars_object` is supposed to be an environment class for now (although we may move to S7 later), can you change it to an environment instead of a list?
Otherwise I think the `$` method would cause strange errors.

etiennebacher: I don't understand, what `polars_object` are you talking about? What should be changed to an environment and would it be a problem to have a list?

eitsupi: Sorry, my mistake!

R/lazyframe-frame.R
```diff
@@ -2312,4 +2312,135 @@ lazyframe__join_asof <- function(
...
+#' # stream a query end-to-end
+#' tmpf2 <- tempfile()
+#' pl$scan_parquet(tmpf)$select(pl$col("cyl") * 2)$sink_parquet(tmpf2)
+#'
+#' # load parquet directly into a DataFrame / memory
+#' pl$scan_parquet(tmpf2)$collect()
+lazyframe__sink_parquet <- function(
+    path,
+    ...,
+    compression = "zstd",
```
eitsupi, line 2371: All allowed values can be listed?

etiennebacher: Done

eitsupi (changes requested):
These tests doesn't seem passable.

etiennebacher:
They now pass.

eitsupi (approved):

PR #122, opened 2025-02-18
feat: Add `rolling()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1316,3 +1316,47 @@ dataframe__partition_by <- function(..., maintain_order = TRUE, include_key = TR
       lapply(\(ptr) .savvy_wrap_PlRDataFrame(ptr) |> wrap())
   })
 }
+
+#' @inherit lazyframe__rolling title description params
+#'
+#' @return [RollingGroupBy][RollingGroupBy_class] (a DataFrame with special
+#' rolling groupby methods like `$agg()`).
```
etiennebacher, line 1323: Currently it automatically opens the docs in `polars` so those should be added in neopolars too

R/dataframe-rolling.R
etiennebacher, line 1: Those are stored in the same file as `group_by` in py-polars (https://github.com/pola-rs/polars/blob/main/py-polars/polars/dataframe/group_by.py), but I think it's clearer to have them in a separate file.

R/lazyframe-frame.R
```diff
@@ -1612,9 +1612,9 @@ lazyframe__unnest <- function(...) {
 #' Create rolling groups based on a date/time or integer column
 #'
 #' @description
-#' Different from `group_by_dynamic`, the windows are now determined by the
+#' Different from `group_by_dynamic()`, the windows are now determined by the
 #' individual values and are not of constant intervals. For constant intervals
-#' use [`<LazyFrame>$group_by_dynamic()`][lazyframe__group_by_dynamic].
+#' use `group_by_dynamic()`.
```
etiennebacher, line 1617: So that we can inherit the description in the DataFrame docs

R/dataframe-frame.R
```diff
@@ -886,7 +886,7 @@ dataframe__gather_every <- function(n, offset = 0) {
 #'
 #' df$rename(foo = "apple")
 # df$rename(
-#   \(column_name) paste0("c", substr(column_name, 2, 100))
+#' #  \(column_name) paste0("c", substr(column_name, 2, 100))
```
etiennebacher, line 889:
```suggestion
#   \(column_name) paste0("c", substr(column_name, 2, 100))
```

R/lazyframe-frame.R
```diff
@@ -1488,7 +1488,7 @@ lazyframe__unpivot <- function(
 #'
 #' lf$rename(foo = "apple")$collect()
 # lf$rename(
-#   \(column_name) paste0("c", substr(column_name, 2, 100))
+#' #  \(column_name) paste0("c", substr(column_name, 2, 100))
```
etiennebacher, line 1491:
```suggestion
#   \(column_name) paste0("c", substr(column_name, 2, 100))
```

R/dataframe-rolling.R
eitsupi: Definitely. This does not necessarily apply for this file, but for other files, the files tend to be too long due to the constraints of having to write the Python methods in a single file. I think there is value in proper division.

However, how about a file name like `dataframe-group_by-rolling` (and rename the original file to `dataframe-group_by-group_by`) to make the relationship with the original file easier to understand?

etiennebacher: > how about a file name like dataframe-group_by-rolling (and rename the original file to dataframe-group_by-group_by) to make the relationship with the original file easier to understand?

Half-agree, I think it's good to have a common "dataframe-group_by" prefix (also when we add group_by_dynamic later) but I see no need in having "dataframe-group_by-group_by". I think that's good enough:

```
dataframe-group_by.R
dataframe-group_by-rolling.R
dataframe-group_by-dynamic.R
```

eitsupi: The reason I want to avoid partial filename matching is because I want to specify the test filename in the `devtools::test()` function's `filter` option.
So how about something like `dataframe-group_by-general.R`? (Rust-side rule)

etiennebacher: Ok sounds good to me.

> I want to specify the test filename in the devtools::test() function's filter option.

I never used it (didn't even know it exists). What's the purpose of this?

PR #127, opened 2025-02-19
refactor: Better error message for `<expr>$reshape()`

etiennebacher:
To rebase on #128


R/expr-expr.R
```diff
@@ -1262,7 +1261,7 @@ expr__dot <- function(other) {
 #' df$select(pl$col("foo")$reshape(c(3, -1)))
 expr__reshape <- function(dimensions) {
   wrap({
-    if (is.numeric(dimensions) && anyNA(dimensions)) {
+    if (anyNA(dimensions)) {
```
etiennebacher, line 1264: Caused the "unreachable" message that was saved in the snapshot

eitsupi: I think we should check the type before `anyNA()`.

eitsupi (changes requested):

R/expr-expr.R
```diff
@@ -1261,7 +1261,13 @@ expr__dot <- function(other) {
 #' df$select(pl$col("foo")$reshape(c(3, -1)))
 expr__reshape <- function(dimensions) {
   wrap({
-    if (is.numeric(dimensions) && anyNA(dimensions)) {
+    if (!is_integerish(dimensions)) {
+      abort("`dimensions` only accepts integer-ish values.")
+    }
+    if (length(dimensions) > 2) {
+      abort("`dimensions` must be of length lower than or equal to 2.")
+    }
```
eitsupi, line 1269:
```suggestion
    if (!is_integerish(dimensions, n = 1L)) {
      abort("`dimensions` only accepts 1 integer-ish value.")
    }
```

eitsupi (changes requested):

R/expr-expr.R
```diff
@@ -1261,7 +1261,10 @@ expr__dot <- function(other) {
 #' df$select(pl$col("foo")$reshape(c(3, -1)))
 expr__reshape <- function(dimensions) {
   wrap({
-    if (is.numeric(dimensions) && anyNA(dimensions)) {
+    if (!is_integerish(dimensions, n = 2)) {
+      abort("`dimensions` only accepts up to 2 integer-ish values.")
+    }
```
eitsupi, line 1266: Sorry but I think I misunderstood, it is past time that reshape did not accept lengths longer than 2, doesn't it work now?

etiennebacher: Oh indeed, I got mistaken because our docs were outdated (actually we have a test with more than 2 dims). I'll fix that.

eitsupi (approved):
Thanks!

PR #129, opened 2025-02-19
fix: Uncomment tests related to `<dataframe>$unnest()` and related fix

etiennebacher:
To rebase on #128


R/expr-expr.R
```diff
@@ -2311,15 +2311,22 @@ expr__hist <- function(
...
     normalize = FALSE) {
   wrap({
     check_dots_empty0(...)
+    if (is.null(name)) {
+      if (isTRUE(normalize)) {
+        name <- "proportion"
+      } else {
+        name <- "count"
+      }
+    }
```
eitsupi, line 2329: roxygen comments also need to be updated?

eitsupi (changes requested):

R/expr-expr.R
etiennebacher: Yes, good catch. Done

eitsupi (approved):

PR #132, opened 2025-02-21
perf: Use custom `$` S3 methods to call Python Polars like methods (except sub namespaces)

eitsupi:
Part of #3

This PR covers objects other than namespaces. Additional work is required for namespaces and will not be done here.


R/datatypes-s3-base.R
```diff
@@ -1,3 +1,65 @@
...
+  member_names <- ls(x)
+  # Enum only method `union`
+  method_names <- c("union", names(polars_datatype__methods))
+
+  all_names <- union(member_names, method_names)
+  filtered_names <- all_names[grep(pattern, all_names)]
+
+  filtered_names[!startsWith(filtered_names, "_")]
+}
+
```
etiennebacher, line 62: I'm confused about this, why do we have a special case for enums?

eitsupi: This is because only Enum has a method called `union`.

etiennebacher (approved):

eitsupi:
@etiennebacher Do you think it is worth pushing this PR further and applying it to all classes before merging? (The rewrite was a far easier task than I thought it would be)

etiennebacher:
Yes I think it's worth it

eitsupi:
Hmmm, I'm not sure why, but I found that running in parallel (i.e. the current default) causes the test to fail.
I may be doing something wrong.

For now, this is a lightweight change on the R side only, so it might be better to leave it and come back to it after setting up the CI.

etiennebacher:
Ok, let's come back to this later

eitsupi:
I found out that the problem that was causing problems in the test (S3 methods such as `print()` and `dim()` were not being called inside some of the testthat functions) was due to the fact that I was exporting custom methods of `.DollarNames` like this:

```
S3method(.DollarNames,polars_dtype)
```

I don't understand the mechanism, but the problem seems to have disappeared after the following modification. (This is the same way to register as in the `R6` package, so perhaps the R6 authors understand this problem)

```
S3method(utils::.DollarNames,polars_dtype)
```

R/expr-s3-base.R
```diff
@@ -1,3 +1,31 @@
...
+  }
+}
+
+#' @exportS3Method utils::.DollarNames
+.DollarNames.polars_expr <- function(x, pattern = "") {
+  member_names <- ls(x, all.names = TRUE)
+  method_names <- names(polars_expr__methods)
+
+  all_names <- union(member_names, method_names)
+  filtered_names <- findMatches(pattern, all_names)
```
etiennebacher, line 24: I've never used `findMatches()` but can you specify `fuzzy = FALSE`?

eitsupi: Just to be sure, why is that?
Not sure which is more convenient.

eitsupi:
Good point.
The current r-polars do it forcefully in files under `R/`, but it is very hard to read the source code.
https://github.com/pola-rs/r-polars/blob/f16e19de23e6a266d08e810f150a1da7c36164dd/R/after-wrappers.R#L141-L154

Perhaps it would be better to generate the R files in a script outside the R package. I will try it later.

eitsupi:
> Just wondering, is it worth creating the `$` and `.DollarNames` methods via an automated script (sort of Rust macros)? There seems to be a lot of duplicated code across all the implementations and there are > 20 methods, so it could help preventing typos or making it easier to modify the methods later.

How about d6131dd?

etiennebacher:
Looks nice, how would that be integrated in the workflow? Would it be an additional step in `task build-documents`?

Nit: why tsv and not csv?

R/dataframe-group_by-rolling.R
```diff
@@ -10,23 +10,17 @@ wrap_to_rolling_group_by <- function(x, index_column, period, offset, closed, gr
...
-    index_column = index_column,
-    period = period,
-    offset = offset,
-    closed = closed,
-    group_by = group_by
+    index_column = self$index_column,
+    period = self$period,
+    offset = self$offset,
+    closed = self$closed,
+    group_by = self$group_by
```
eitsupi, line 23: This was an original implementation error, but it seemed to work. It needed to be fixed because it stopped working here.

eitsupi:
> why tsv and not csv?

It may be easier to read depending on how it is displayed, and I feel that it is slightly more reliable than csv because tabs are more likely to appear in a value than commas.

R/datatypes-s3-base.R
```diff
@@ -1,3 +1,37 @@
+#' @export
+`$.polars_dtype_enum` <- function(x, name) {
+  # Enum only method `union`
```
etiennebacher, line 3: Can you add a comment explaining this is why it's not in `generate-S3methods-dollar.R`?

dev/generate-r-files/data/class-names.tsv
```diff
@@ -0,0 +1,9 @@
+class_name	env_name
+polars_data_frame	polars_dataframe__methods
+polars_group_by	polars_groupby__methods
+polars_rolling_group_by	polars_rolling_groupby__methods
+polars_dtype	polars_datatype__methods
```
etiennebacher, line 5: Not specifically related to this PR but shouldn't this be named `polars_dtype_methods` for consistency?

Taskfile.yml
```diff
@@ -187,11 +199,13 @@ tasks:
       # These files are auto-generated and ignored by the formatter
       - exclude: R/000-wrappers.R
       - exclude: R/import-standalone-*.R
+      - exclude: R/generated-*.R
```
etiennebacher, line 202: IIUC this is to avoid invalidating the task, but shouldn't this also be added in `air.toml`?

etiennebacher (commented):
Thanks, looks good to me, just a few comments.

I think the CI workflows should also be updated to run when the devfiles are modified. Also, the expr subnamespaces are not handled here?

eitsupi:
> I think the CI workflows should also be updated to run when the devfiles are modified.

Which workflow?

Taskfile.yml
eitsupi: It is assumed that Air ignores these files because of the `# Generated by` comment on the first line.

etiennebacher:
Ah no I meant when the R CMD check workflow is implemented.

eitsupi:
Sorry, I have not yet implemented the R CMD check workflow because I have not yet written the configure scripts, etc.

> Also, the expr subnamespaces are not handled here?

I have updated the PR description, and will tackle these later, as I have found that they require additional work.

PR #134, opened 2025-02-22
feat: Add `group_by_dynamic()` for DataFrame

etiennebacher:


R/lazyframe-frame.R
```diff
@@ -1841,7 +1842,11 @@ lazyframe__group_by_dynamic <- function(
     every <- parse_as_duration_string(every)
     offset <- parse_as_duration_string(offset) %||% "0ns"
     period <- parse_as_duration_string(period) %||% every
-    group_by <- parse_into_list_of_expressions(!!!group_by)
+    if (!is_polars_expr(group_by)) {
+      group_by <- parse_into_list_of_expressions(!!!group_by)
+    } else {
+      group_by <- list(group_by$`_rexpr`)
+    }
```
etiennebacher, line 1849: `parse_into_list_of_expressions()` fails when `group_by` is an expression. It doesn't look great but I can't remember if we had cases like this already

R/lazyframe-frame.R
```diff
@@ -1749,7 +1749,7 @@ lazyframe__rolling <- function(
 #' - 1i # length 1
 #' - 10i # length 10
 #'
-#' @return A [LazyGroupBy][LazyGroupBy_class] object
+#' @return A [LazyGroupByDynamic][LazyGroupByDynamic_class] object
```
etiennebacher, line 1752: Not sure if we should have the same class for eager and lazy

R/lazyframe-frame.R
eitsupi: An oversight in my review in #49: there are problems with the implementation of `lazyframe__rolling` and `lazyframe__group_by_dynamic` and we need to use dynamic-dots instead of the `group_by` argument.

R/lazyframe-frame.R
eitsupi: IIUC, these are not distinguished on the LazyFrame (i.e. Rust side), so it is not worth distinguishing them on the Python/R side.

R/lazyframe-frame.R
etiennebacher: While it's ok to use `...` in `group_by()` because there is no other column selection, the `group_by` argument in `rolling()` and `group_by_dynamic()` is "just" an extra argument and isn't central to using those functions. Therefore, capturing all dots and assign them to `group_by` isn't ideal IMO. It also means that we wouldn't error on misnamed arguments.

Wouldn't it better to accepts expressions in `parse_into_list_of_expressions()` instead, as in py-polars?

Either way, `rolling()` needs to be adjusted.

R/lazyframe-frame.R
etiennebacher: I've put `GroupByDynamic` in `@return` for eager and lazy

R/lazyframe-frame.R
eitsupi: > Wouldn't it better to accepts expressions in `parse_into_list_of_expressions()` instead, as in py-polars?

Since this function here is intended to be passed directly to `...` for ease of use, I think it needs to be processed before passing it to this function rather than having an exception inside this function.

eitsupi: Oh, sorry. I think I misunderstood the role of these arguments.
I think this fix should be fine, but I'll look at it some more later.

eitsupi (approved):
Thanks!

PR #135, opened 2025-02-22
build: Add settings for the Air formatter for R files

eitsupi:
Related to #73

The actual formatting has not yet been done. (I would like to do this after this is merged.)

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -11,7 +11,6 @@ vars:
   MANIFEST: src/rust/Cargo.toml
   CARGO_LOCK: src/rust/Cargo.lock
   R_SOURCE: R/*
-  R_TESTS: tests/*
   VIGNETTES: vignettes/**/*.Rmd
   RUST_SOURCE: src/rust/src/**/*.rs
 
@@ -88,7 +87,6 @@ tasks:
     deps:
       - build-rust
       - build-standalone-files
-      - format-r
     cmds:
       - Rscript -e
         '{{.R_FUNC_LOAD_REQUIRED_PKGS}};
@@ -160,22 +158,3 @@ tasks:
     desc: Accept all test snapshots. (Shortcut to accept snapshots after running tests)
     cmds:
       - Rscript -e 'testthat::snapshot_accept()'
-
-  format-r:
-    desc: Format R files.
-    aliases:
-      - fmt-r
-    sources:
-      - "{{.R_SOURCE}}"
-      - "{{.R_TESTS}}"
-      - air.toml
-      # These files are auto-generated and ignored by the formatter
-      - exclude: R/000-wrappers.R
-      - exclude: R/import-standalone-*.R
-    generates:
-      - "{{.R_SOURCE}}"
-      - "{{.R_TESTS}}"
-      - exclude: R/000-wrappers.R
-      - exclude: R/import-standalone-*.R
-    cmds:
-      - air format R/ tests/
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -11,6 +11,7 @@ vars:
   MANIFEST: src/rust/Cargo.toml
   CARGO_LOCK: src/rust/Cargo.lock
   R_SOURCE: R/*
+  R_TESTS: tests/*
   VIGNETTES: vignettes/**/*.Rmd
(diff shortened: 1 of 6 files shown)


air.toml
```diff
@@ -0,0 +1,6 @@
+[format]
+line-width = 120
```
eitsupi, line 2: I set it at 120, the same as the current r-polars, because I feel that 80 is too short in this day and age, but 120 may be too long.
https://github.com/pola-rs/r-polars/blob/d0e1d6a31d7f0c624e52ee316c861e6595e4ee60/.lintr.R#L2

@etiennebacher Thoughts?

etiennebacher: I think 80 (or close to 80) is recommended or the default in several other places:

* [tidyverse style guide](https://style.tidyverse.org/syntax.html#long-function-calls)
* Black recommends [88 characters](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)
* I think Ruff also enforces 80 by default but I can't find a link

120 is definitely too long for me, especially since I'm working on laptop these days. I'd be willing to try 90 or 100 max.

eitsupi: Ok, the Linux kernel has relaxed the limit from 80 to 100. 100 looks good.

air.toml
```diff
@@ -0,0 +1,13 @@
...
+indent-style = "space"
+line-ending = "auto"
+persistent-line-breaks = true
+exclude = [
+    # Syntax error bug of the formatter
+    # https://github.com/posit-dev/air/issues/255
+    "tests/testthat/test-as_polars_series.R",
+    "tests/testthat/test-expr-string.R",
+    "tests/testthat/test-datatypes-classes.R",
+]
```
etiennebacher, line 13: I'm confused, do you want to ignore those or remove raw strings as said in https://github.com/eitsupi/neo-r-polars/pull/73#issuecomment-2676138156?

eitsupi: Sorry for the confusion. My intention was to ensure that the `task` command would succeed at this PR stage.

We can update the R files later to remove these lines, so I will add a TODO comment.

etiennebacher: Actually I had a patch ready for this in #73 so I can push it there and you can remove your lines here.

Taskfile.yml
```diff
@@ -158,3 +160,22 @@ tasks:
...
+      # These files are auto-generated and ignored by the formatter
+      - exclude: R/000-wrappers.R
+      - exclude: R/import-standalone-*.R
+    generates:
+      - "{{.R_SOURCE}}"
+      - "{{.R_TESTS}}"
+      - exclude: R/000-wrappers.R
+      - exclude: R/import-standalone-*.R
+    cmds:
+      - air format R/ tests/
```
etiennebacher, line 181: Can we put this in #73 instead?

I don't understand this PR, I thought it was just to set `air.toml`.

eitsupi: I was checking the behavior here because these have interactions; it's OK to have them cherry-picked at #73 and removed here.

eitsupi:
@etiennebacher Does this look OK?

air.toml
```diff
@@ -0,0 +1,14 @@
+[format]
+line-width = 100
+indent-width = 2
+indent-style = "space"
+line-ending = "auto"
+persistent-line-breaks = true
+exclude = [
```
etiennebacher, line 7: This can be removed due to [`bf5a253` (#73)](https://github.com/eitsupi/neo-r-polars/pull/73/commits/bf5a253c1e789fa8019d22ad6c73f33535f991b3)

etiennebacher (approved):
Just `exclude` to remove, otherwise looks good, thanks

air.toml
```diff
@@ -0,0 +1,6 @@
+[format]
+line-width = 100
+indent-width = 2
+indent-style = "space"
+line-ending = "auto"
```
eitsupi, line 5: My bad, this should be `"lf"` as like `.editorconfig`.
https://github.com/eitsupi/neo-r-polars/blob/59ef78d92120f9187748b8a10f95b79ecdcef53d/.editorconfig#L4

eitsupi: I opened #139

PR #136, opened 2025-02-22
style: Formatting to make the Air formatter happy

eitsupi:


eitsupi:
@etiennebacher Could you take a look at this?

R/config.R
```diff
@@ -1,8 +1,12 @@
 parse_env_auto_structify <- function() {
   value <- Sys.getenv("POLARS_AUTO_STRUCTIFY", "0")
-  switch(value,
+  switch(
+    value,
```
etiennebacher, line 4: Need to skip this one?

etiennebacher (approved):
Thanks, LGTM, but I wonder if we want to skip the `switch()` calls. I don't think the autoformatting output is bad on those calls and the fewer `fmt: skip` comments the better.

eitsupi:
Thanks for looking. I probably don't understand how Air feels.
I added the comment because when I first formatted it, it was very difficult to read because of the long horizontal formatting within the switch.
Like:

```r
switch(foo, bar = 1, baz = 2, ...)
```

Glad I could delete these 'fmt: skip' comments later.

etiennebacher:
>  I added the comment because when I first formatted it, it was very difficult to read because of the long horizontal formatting within the switch.

Ah ok, I thought it would always reformat as
```r
switch(
  foo,
  a = 1,
  ...
)
```
Let's leave the comments like this then.

PR #142, opened 2025-02-23
build(taskfile): Add aliases for `build-rust` and `build-documents`

etiennebacher:
Since we added aliases for formatting, those are the two I use the most.


eitsupi (approved):
LGTM.

Just to be sure, if you use the task command a lot, autocomplete is available in the major Shells.
https://taskfile.dev/installation/#setup-completions

PR #143, opened 2025-02-24
feat: Add `describe()` for DataFrame and LazyFrame

etiennebacher:


R/datatypes-classes.R
```diff
@@ -302,3 +302,19 @@ datatype__min <- function() {
...
+  inherits(self, "polars_dtype_categorical")
+}
+
+datatype__is_enum <- function() {
+  inherits(self, "polars_dtype_enum")
+}
+
+datatype__is_null <- function() {
+  inherits(self, "polars_dtype_null")
+}
```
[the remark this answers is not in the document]
etiennebacher: Needed in the `skip_minmax()` in `lazyframe__describe`

R/lazyframe-frame.R
```diff
@@ -2610,4 +2610,185 @@ lazyframe__sink_csv <- function(
...
+    # create list of metrics
+    metrics <- c("count", "null_count", "mean", "std", "min")
+    quantiles <- parse_percentiles(percentiles)
+    if (length(quantiles) > 0) {
+      metrics <- c(metrics, paste0(percentiles * 100, "%"))
+    }
+    metrics <- c(metrics, "max")
+
+    skip_minmax <- function(x) {
+      # TODO: need to check if x is object, or unknown
```
etiennebacher, line 2671: Can we have a datatype object or unknown?

R/datatypes-classes.R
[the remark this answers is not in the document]
eitsupi: Is it really so?　These should not exist in Python Polars either.

R/lazyframe-frame.R
eitsupi: All types are listed here; Unknown exists but Object does not.

https://github.com/eitsupi/neo-r-polars/blob/c053c0f9f8677c67356839bf4bebb7156747548d/src/rust/src/datatypes.rs#L263-L293

R/datatypes-classes.R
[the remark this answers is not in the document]
etiennebacher: They do it like this in `describe()`:

```
@lru_cache
def skip_minmax(dt: PolarsDataType) -> bool:
     return dt.is_nested() or dt in (Categorical, Enum, Null, Object, Unknown)
```
but AFAIK we don't really have sth equivalent.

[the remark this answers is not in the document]
eitsupi: Something like:

```r
skip_minmax <- function(dtype) {
  dtypes <- class(dtype)

  dtype$is_nested() ||
    any(
      dtypes[1] ==
        c(
          "polars_dtype_categorical",
          "polars_dtype_enum",
          "polars_dtype_null",
          "polars_dtype_unknown"
        )
    )
}
```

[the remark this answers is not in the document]
etiennebacher: Ah indeed, we can even make it more robust with 
```
inherits(dtype, c(.... )) 
```
since `inherits()` returns TRUE if any of the classes is matched

R/lazyframe-frame.R
eitsupi: > we have any examples with this.

Indeed, there is no exposed way to create an instance of this type.

```r
dt <- PlRDataType$new_from_name("Unknown") |> wrap()
dt
#> Unknown(Any)
class(dt)
#> [1] "polars_dtype_unknown" "polars_dtype"         "polars_object" 

```

R/lazyframe-frame.R
```diff
@@ -2610,4 +2610,199 @@ lazyframe__sink_csv <- function(
...
+    # Cast by column type (numeric/bool -> float), (other -> string)
+    # This is done later in py-polars but we need to do it here. When we gather
+    # values in the list, we coerce the types, meaning that stats for date (for
+    # instance) are coerced to numeric. Casting those to string beforehand
+    # fixes this.
+    df_metrics_schema <- df_metrics$schema
+    for (i in seq_along(df_metrics_schema)) {
+      nm <- names(df_metrics_schema)[i]
+      dtype <- df_metrics_schema[[i]]
+      df_metrics_dtype <- if (dtype$is_numeric() || inherits(dtype, "polars_dtype_boolean")) {
```
etiennebacher, line 2782: I still think it would be nice to have `is_boolean()` and friends because this is kind of weird. Problem for later

eitsupi: I recall pola-rs/polars#19235

etiennebacher: Yes, too bad they don't want to implement those

etiennebacher:
Is this ok to get merged?

eitsupi (approved):
Thanks, I was hoping to refactor it but I don't have time so I guess we can merge it for now and do it later.

R/lazyframe-frame.R
```diff
@@ -2361,4 +2361,199 @@ lazyframe__join_asof <- function(
...
+#'
+#' # Customize which percentiles are displayed, applying linear interpolation:
+#' lf$describe(
+#'   percentiles = c(0.1, 0.3, 0.5, 0.7, 0.9),
+#'   interpolation = "linear"
+#' )
+lazyframe__describe <- function(
+  percentiles = c(0.25, 0.5, 0.75),
+  interpolation = c("nearest", "higher", "lower", "midpoint", "linear")
+) {
```
eitsupi, line 2406: In Python Polars, `interpolation` must be named.
So we need `...` and check dots empty.

eitsupi (changes requested):

R/lazyframe-frame.R
etiennebacher: Good catch, done

eitsupi (approved):

PR #147, opened 2025-02-24
feat: Add `write_parquet()` for DataFrame

etiennebacher:
- [ ] rebase on #121 
- [ ] uncomment examples requiring `write_parquet()`


src/rust/src/dataframe/general.rs
```diff
@@ -304,4 +306,95 @@ impl PlRDataFrame {
     pub fn is_empty(&self) -> Result<Sexp> {
         self.df.is_empty().try_into()
     }
+
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn write_parquet(
```
eitsupi, line 311: Do not put this function here; check the source code of polars-python.

eitsupi (changes requested):

src/rust/src/dataframe/io.rs
```diff
@@ -38,4 +39,95 @@ impl PlRDataFrame {
             .map_err(RPolarsErr::from)?;
         Ok(df.into())
     }
+
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn write_parquet(
```
eitsupi, line 44: As I mentioned somewhere before, I am wondering if it is worthwhile to add this now, given that we have problems with this (pola-rs/polars#13040)

etiennebacher: I think this is an important function. There's no indication that the linked issue will be resolved any time soon so I don't think we should wait for it.

Also `write_parquet()` is already supported in `r-polars`.

eitsupi: Sorry, I was confused. I think the above issue is related to reading and there are two write functions, which is not a problem.

tests/testthat/test-io-parquet-functions.R
```diff
@@ -138,132 +138,140 @@ test_that("scan_parquet can include file path", {
...
+  dat <- as_polars_df(mtcars)
+  tmpf <- withr::local_tempfile()
+  x <- dat$write_parquet(tmpf)
+  expect_equal(x, dat)
+})
+
+test_that("write_parquet: argument 'statistics'", {
+  dat <- as_polars_df(mtcars)
+  tmpf <- withr::local_tempfile()
+  on.exit(unlink(tmpf))
```
eitsupi, line 174: IIUC, the file created by ` withr::local_tempfile` will be removed automatically, so we don't need this.

eitsupi (approved):
Let's merge for now, as I think we can fix the tests and such later.
Thanks!

PR #148, opened 2025-02-24
feat: Add `sink_ipc()` for LazyFrame

etiennebacher:


src/rust/src/conversion/mod.rs
```diff
@@ -862,3 +862,17 @@ pub(crate) fn parse_parquet_compression(
     };
     Ok(parsed)
 }
+
+#[cfg(not(target_arch = "wasm32"))]
```
eitsupi, line 866: Is this needed?

src/rust/src/lazyframe/general.rs
```diff
@@ -1263,4 +1263,50 @@ impl PlRLazyFrame {
             .map_err(RPolarsErr::from);
         Ok(())
     }
+
+    #[cfg(not(target_arch = "wasm32"))]
```
eitsupi, line 1267: Ditto.

R/lazyframe-frame.R
```diff
@@ -2485,4 +2485,65 @@ lazyframe__sink_parquet <- function(
...
+  storage_options = NULL,
+  retries = 2
+) {
+  wrap({
+    check_dots_empty0(...)
+    compression <- compression %||% "uncompressed"
+    compression <- arg_match0(
+      compression,
+      values = c("lz4", "uncompressed", "zstd")
+    )
```
eitsupi, line 2523: Does not match to the Rust side.

eitsupi (changes requested):

src/rust/src/conversion/mod.rs
etiennebacher: No, I forgot that wasm supports sinks, just not parquet

R/lazyframe-frame.R
```diff
@@ -2485,4 +2485,63 @@ lazyframe__sink_parquet <- function(
...
 
+#' Evaluate the query in streaming mode and write to an IPC file
+#'
+#' @inherit lazyframe__sink_parquet description params return
+#' @inheritParams rlang::args_dots_empty
+#' @param compression Must be one of:
+#' * `"lz4"`: fast compression/decompression.
+#' * `"zstd"`: good compression performance.
+#'
+#' If `NULL`, it uses `"zstd"`.
```
eitsupi, line 2496: I think this interpretation is incorrect: the official JS implementation of Apache Arrow cannot read compressed Arrow files. Therefore, uncompressed Arrow files are important.
If `NULL` is interpreted as `"zstd"`, it means that there is no way to write an uncompressed Arrow file.

See also pola-rs/polars#15644

eitsupi (changes requested):

R/lazyframe-frame.R
eitsupi: I was going to fix this as it is complicated to explain, but I didn't have time today.

eitsupi: In short, the mismatch with the Parquet writer on the Rust side requires a different strategy.

etiennebacher: I moved the function and tests to dedicated files so the comments here are marked outdated but are still relevant.

I'm still puzzled by this, upstream polars uses allows `"uncompressed"` on the python side but this hits the `must be one of {'zstd', 'lz4'}` error when passed to Rust. Can you finish this PR since you seem to know what to do?

etiennebacher:
https://github.com/pola-rs/polars/pull/21786

etiennebacher:
The PR linked in the previous comment made equivalent `compression = NULL` and `compression = "uncompressed"`. I don't really understand where in upstream code but `py-polars` passes `"uncompressed"` to rust-polars and it is converted to `None`. Here we simply convert `"uncompressed"` to `NULL` in R and pass `Option<&str>` to Rust.

R/output-ipc-functions.R
```diff
@@ -1 +1,64 @@
...
+  storage_options = NULL,
+  retries = 2
+) {
+  wrap({
+    check_dots_empty0(...)
+    compression <- compression %||% "uncompressed"
+    compression <- arg_match0(compression, values = c("zstd", "lz4", "uncompressed"))
+    if (compression == "uncompressed") {
+      compression <- NULL
+    }
```
etiennebacher, line 40: Convert `NULL` to `"uncompressed"` and then back to `NULL` so that `arg_match()` doesn't fail when it receives `NULL`.

eitsupi (approved):
Thanks!

PR #155, opened 2025-02-26
fix: Fix `<expr>$map_batches()` crashing R session

### LINKED ISSUE #5: `<expr>$map_batches()` may crash the session
It seems that if we run `map_batches()` consecutively as shown below, an error will eventually occur.

```r
while(TRUE) {pl$select(pl$lit("a")$map_batches(\(...) 1)) |> print()}
```

```
Error: bad value
Error:
Caused by error:
Caused by error in `PlRLazyFrame_filter()`:
! invalid formal argument list for "function"
Run `rlang::last_trace()` to see where the error occurred.
```

I think this indicates that something is wrong with inter-thread communication.
The same error occurs for the current r-polars, but it does not appear to occur when I set `in_background=TRUE` to have the UDF evaluated in a separate R session.

### LINKED ISSUE #23: `<lazyframe>$group_by()$agg()$collect()` may hang
I noticed this example will hang.

```r
as_polars_df(mtcars)$group_by("cyl", foo = TRUE)$agg()
```

This is not confirmed by Python Polars or current R Polars.
It may occur if some feature flag is missing, but I have not been able to identify it (Enabling several features did not solve the problem.).

eitsupi:
Fix #5

I definitely misunderstood the protection and release of R objects.
I don't think there is any need for special protection on the Rust side since the function object is copied when it is passed.

### DIFF
--- a/src/rust/src/r_udf.rs
+++ b/src/rust/src/r_udf.rs
@@ -2,43 +2,30 @@
 // In the case of Python, it does not exist because there is a special feature in polars-plan.
 
 use crate::{prelude::*, r_threads::ThreadCom, PlRSeries};
-use savvy::{
-    ffi::SEXP,
-    protect::{insert_to_preserved_list, release_from_preserved_list},
-    FunctionArgs, FunctionSexp, Sexp, TypedSexp,
-};
+use savvy::{FunctionArgs, FunctionSexp, Sexp, TypedSexp};
 use state::InitCell;
-use std::sync::{Arc, Mutex, RwLock};
-
-// TODO: comments
-#[derive(Clone)]
-struct UnsafeToken(SEXP);
-unsafe impl Send for UnsafeToken {}
-unsafe impl Sync for UnsafeToken {}
+use std::sync::RwLock;
 
 struct UnsafeFunctionSexp(FunctionSexp);
 unsafe impl Send for UnsafeFunctionSexp {}
 unsafe impl Sync for UnsafeFunctionSexp {}
 
+impl Clone for UnsafeFunctionSexp {
+    fn clone(&self) -> Self {
+        Self(FunctionSexp(self.0.inner().clone()))
+    }
+}
+
 #[derive(Clone)]
 pub struct RUdf {
-    function: Arc<Mutex<UnsafeFunctionSexp>>,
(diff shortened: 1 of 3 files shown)


eitsupi:
@etiennebacher Could you try this?

etiennebacher:
Is it really for #23? I just don't see any call to map_batches() in the example there so just checking it's not a typo.

```r
as_polars_df(mtcars)$group_by("cyl", foo = TRUE)$agg()
```
panics for me:
```
panic occured!

Original message:
    panicked at /home/etienne/.cargo/git/checkouts/polars-54176cdfb679e240/ff994f0/crates/polars-core/src/frame/mod.rs:523:17:
    assertion `left == right` failed
      left: 0
     right: 16

Backtrace:
       0: savvy::panic_hook::get_backtrace
                 at 
[....]
```

etiennebacher:
If it is for #5 instead, then the example there also instantly panics:

```r
x <- 0
while(TRUE) {pl$select(pl$lit("a")$map_batches(\(...) 1)); print(x <<- x + 1)}
```
```
panic occured!

Original message:
    panicked at src/r_threads.rs:53:14:
    thread failed send, likely a user interrupt: "SendError(..)"

Backtrace:
       0: savvy::panic_hook::get_backtrace
                 at 
```

eitsupi:
Thanks for taking a look at this.

Sorry, I made a mistake on the referenced issue. As far as I can tell on Linux, the error does not occur, but the problem may not be completely resolved (or a different issue).
I'll merge it for now, but leave the test as skip.

etiennebacher:
> As far as I can tell on Linux, the error does not occur, but the problem may not be completely resolved (or a different issue).

I'm currently on ubuntu and it still appears

etiennebacher:
Uh my bad, now it works. Maybe an unclean package compilation on my side. 5k iterations and going

etiennebacher:
Do you want to remove the TODOs related to `map_batches()` in R files and remove the `@noRd` for its docs?

eitsupi:
Thanks confirming!

Now let's revive the documentation and tests.

eitsupi:
Oh, sorry. The documentation seems to have been copied from the current r-polars and is not accurate. I'll add it later.

PR #156, opened 2025-02-26
docs: Update documentation of `map_batches()`

etiennebacher:
I thought `returns_scalar` and `is_elementwise` were already used but they require more work on the Rust side.


R/expr-expr.R
```diff
@@ -1070,102 +1070,117 @@ expr__filter <- function(...) {
...
+#' scalar that will be converted into a Series. If the result is a scalar and
+#' you want it to stay as a scalar, pass in `returns_scalar = TRUE`.
 #'
+#' If you want to apply a custom function elementwise over single values, see
+#' [map_elements()][expr__map_elements]. A reasonable use case for map
+#' functions is transforming the values represented by an expression using a
+#' third-party package.
+#'
+#' @inheritParams rlang::args_dots_empty
+#' @param f Function to apply.
```
etiennebacher, line 1086: Param named `function` in py-polars but it's a keyword in R

R/expr-expr.R
```diff
@@ -1070,102 +1070,117 @@ expr__filter <- function(...) {
...
+#'   a = c(0, 1, 0, 1),
+#'   b = c(1, 2, 3, 4)
+#' )
+#' system.time({
+#'   print(
+#'     df$group_by("a")$agg(
+#'       pl$col("b")$map_batches(\(x) x + 2, agg_list = FALSE)
+#'     )
+#'   )
+#' })
```
etiennebacher, line 1127: This is actually faster than the alternative below so far, it might change with the rework on `.DollarNames`

src/rust/src/map/lazy.rs
```diff
@@ -9,13 +9,13 @@ use savvy::{FunctionSexp, Result};
...
     lambda: FunctionSexp,
-    output_type: Option<&PlRDataType>,
+    return_dtype: Option<&PlRDataType>,
     agg_list: bool,
-    // TODO: support these options
-    // is_elementwise: bool,
-    // returns_scalar: bool,
+    // TODO: use these args
+    _is_elementwise: bool,
+    _returns_scalar: bool,
```
eitsupi, line 16: Why this change?

eitsupi (changes requested):

eitsupi (changes requested):
Not only this time, but please stop making significant non-document changes in the `doc` type.

R/expr-expr.R
etiennebacher: I renamed to `lambda`, it's not like R is consistent on this (sometimes `FUN`, sometimes `f` for function, sometimes `f` for factors, etc.)

etiennebacher:
Done, it's true I tend to mix a bit of features/fixes/docs, sorry about that

eitsupi (approved):
Thanks, let's merge for now.

PR #158, opened 2025-02-27
fix: Fix R CMD check `NOTE`s in `fill_null()` for LazyFrame

etiennebacher:


DESCRIPTION
```diff
@@ -16,6 +16,7 @@ License: MIT + file LICENSE
 Depends:
     R (>= 4.2)
 Imports:
+    methods,
```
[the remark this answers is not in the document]
etiennebacher: Actually I used `@importFrom`

eitsupi (commented):
Can you use `infer_polars_dtype()` instead of check R classes?

R/expr-expr.R
```diff
@@ -3914,7 +3914,7 @@ expr__fill_null <- function(value, strategy = NULL, limit = NULL) {
       abort("can only specify `limit` when strategy is set to 'backward' or 'forward'")
     }
     if (!missing(value)) {
-      self$`_rexpr`$fill_null(as_polars_expr(value)$`_rexpr`)
+      self$`_rexpr`$fill_null(as_polars_expr(value, as_lit = TRUE)$`_rexpr`)
```
etiennebacher, line 3917: Otherwise tests on LazyFrame `fill_null()` fail.

tests/testthat/test-lazyframe-frame.R
```diff
@@ -797,6 +797,26 @@ test_that("fill_null(): basic usage", {
...
+  expect_query_equal(
+    .input$fill_null("foobar"),
+    df,
+    pl$DataFrame(
+      num = c(1, 2, NA),
+      str = c("a", "b", "foobar"),
+      cat_lex = c("a", "b", "foobar"),
+      cat_phy = c("a", "b", "foobar")
+    )$cast(cat_lex = pl$Categorical("lexical"), cat_phy = pl$Categorical("physical"))
+  )
```
etiennebacher, line 817: This throws a message about how categorical remapping is expensive (also in py-polars), but it doesn't fail the tests.

etiennebacher:
Done, nice simplification indeed

R/lazyframe-frame.R
```diff
@@ -1006,11 +1006,10 @@ lazyframe__fill_null <- function(
     }
     check_dots_empty0(...)
     if (!missing(value) && !is.null(value)) {
+      dtype <- infer_polars_dtype(value)
```
eitsupi, line 1009: I think this would be an error if we passed a polars_expr for `value`. Test cases are needed.

eitsupi (changes requested):

R/expr-expr.R
```diff
@@ -3904,7 +3904,9 @@ expr__fill_nan <- function(value) {
 #' )
 expr__fill_null <- function(value, strategy = NULL, limit = NULL) {
   wrap({
-    check_exclusive(value, strategy)
+    if (!missing(value) && !missing(strategy) && !is.null(value) && !is.null(strategy)) {
+      abort("Exactly one of `value` or `strategy` must be supplied.")
+    }
```
etiennebacher, line 3909: This is necessary because when we call this `fill_null()` from the LazyFrame method, `value` is something and `strategy` is `NULL`, which fails this check.

It is annoying that `check_exclusive()` fails if both values are `NULL` or one is something and the other is `NULL`, how about we make a variant of this function that allows those cases?

eitsupi: I think that makes sense.

R/lazyframe-frame.R
```diff
@@ -1002,41 +1002,31 @@ lazyframe__fill_null <- function(
...
+        } else if (inherits(dtype, "polars_dtype_string")) {
+          dtypes <- c(pl$String, pl$Categorical("physical"), pl$Categorical("lexical"))
+        } else {
+          dtypes <- dtype
+        }
       }
-      # TODO: time datatype
 
-      if (!is_list_of_polars_dtype(dtypes)) {
+      if (!is.null(dtypes) && !is_list_of_polars_dtype(dtypes)) {
```
eitsupi, line 1029: Maybe `if (is_polars_dtype(dtypes))`?

R/expr-expr.R
etiennebacher: Done in [`d9c70ed`](https://github.com/eitsupi/neo-r-polars/pull/158/commits/d9c70edf58a563e3d6f153a7ba0439290ababa7e)

R/lazyframe-frame.R
etiennebacher: Done (and added test for date and datetime that should cover this case)

tests/testthat/test-lazyframe-frame.R
```diff
@@ -797,6 +805,51 @@ test_that("fill_null(): basic usage", {
...
+    dt = as.Date(c("2020-01-01", NA)),
+    dtime = as.POSIXct(c("2020-01-01 00:00:00", NA))
+  )
+
+  expect_query_equal(
+    .input$fill_null(as.Date("2020-01-02")),
+    df,
+    pl$DataFrame(
+      dt = as.Date(c("2020-01-01", "2020-01-02")),
+      dtime = as.POSIXct(c("2020-01-01 00:00:00", NA))
```
eitsupi, line 839: Ditto.

eitsupi (changes requested):
Thanks, I think one more test update is needed, but looks good.

tests/testthat/test-lazyframe-frame.R
```diff
@@ -797,6 +805,51 @@ test_that("fill_null(): basic usage", {
...
+      cat_lex = c("a", "b", "foobar"),
+      cat_phy = c("a", "b", "foobar")
+    )$cast(cat_lex = pl$Categorical("lexical"), cat_phy = pl$Categorical("physical"))
+  )
+})
+
+test_that("fill_null() works on date/datetime", {
+  df <- pl$DataFrame(
+    dt = as.Date(c("2020-01-01", NA)),
+    dtime = as.POSIXct(c("2020-01-01 00:00:00", NA))
```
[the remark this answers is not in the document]
etiennebacher: Thanks, done.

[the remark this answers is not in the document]
eitsupi: Thanks for the update.
In fact I always use the UTC time zone for safety, so I never run into the bug myself.
I was simply pointing out best practices.

eitsupi (approved):
Thanks!
(Never mind the comments from Copilot. I just recently found out about this feature and wanted to give it a try.)

PR #162, opened 2025-02-27
feat: Add `write_csv()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1888,3 +1888,65 @@ dataframe__write_parquet <- function(
...
+#' @inherit dataframe__write_parquet return
+#' @examples
+#' tmpf <- tempfile()
+#' as_polars_df(mtcars)$write_csv(tmpf)
+#' pl$read_csv(tmpf)
+#'
+#' as_polars_df(mtcars)$write_csv(tmpf, separator = "|")
+#' pl$read_csv(tmpf, separator = "|")
+dataframe__write_csv <- function(
+  file,
```
etiennebacher, line 1906: In py-polars, they have the option of this being `None` to return the CSV content as a string, but I'm not sure how to do this here (or this is actually useful)

R/lazyframe-frame.R
```diff
@@ -2552,7 +2552,7 @@ lazyframe__sink_csv <- function(
   float_scientific = NULL,
   float_precision = NULL,
   null_value = "",
-  quote_style = "necessary",
+  quote_style = c("necessary", "always", "never", "non_numeric"),
```
etiennebacher, line 2555: Drive-by, forgot that in sink_csv PR

src/rust/src/dataframe/io.rs
```diff
@@ -130,4 +131,81 @@ impl PlRDataFrame {
...
+            }
+            None => None,
+        };
+        let cloud_options = {
+            let cloud_options =
+                parse_cloud_options(path.to_str().unwrap(), cloud_options.unwrap_or_default())?;
+            Some(cloud_options.with_max_retries(retries))
+        };
+
+        let f = std::fs::File::create(path).map_err(RPolarsErr::from)?;
```
etiennebacher, line 189: This is used here in py-polars: https://github.com/pola-rs/polars/blob/e801b835b7951c32c504553ebc5b7479eb66625d/crates/polars-python/src/dataframe/io.rs#L375

But we don't have the equivalent, we don't support cloud writing anywhere AFAIK.

tests/testthat/_snaps/io-csv-functions.md
```diff
@@ -1,4 +1,4 @@
-# arg raise_if_empty works
+# read/scan: arg raise_if_empty works
```
etiennebacher, line 1: I put the tests in io-csv-functions, where there are already the scan/read functionalities so I added this to separate the read/scan and the write tests.

On the other hand, tests for sink_csv are in test-lazyframe-frame.R. Maybe we could have "test-csv-read-scan.R" and test-csv-sink-write.R"?

tests/testthat/test-io-csv-functions.R
```diff
@@ -296,3 +298,184 @@ test_that("arg 'decimal_comma' works", {
     pl$DataFrame(a = c(1.5, 2), b = c("a", NA), c = c(2L, NA))$cast(c = pl$Int64)
   )
 })
+
+# write ------------------------------------------------
+
+test_that("write_csv works", {
```
etiennebacher, line 304: Basically all the tests of sink_csv() but adapted for write_csv()

eitsupi (approved):
Thanks!

PR #163, opened 2025-02-27
docs: Several fixes for forgotten args and wrong links

etiennebacher:


R/expr-string.R
```diff
@@ -24,7 +24,7 @@ namespace_expr_str <- function(x) {
 #' Similar to the [strptime()] function.
 #'
 #' When parsing a Datetime the column precision will be inferred from the format
-#' string, if given, e.g.: `"%F %T%.3f"` => [`pl$Datetime("ms")`][pl_Datetime].
+#' string, if given, e.g.: `"%F %T%.3f"` => [`pl$Datetime("ms")`][pl__datetime].
```
eitsupi, line 27: I suppose `pl__Datetime` instead of `pl__datetime`.

https://github.com/eitsupi/neo-r-polars/blob/e2cfea96b49f5426219b57aa837ff7ab99e862b4/R/datatypes-classes.R#L177

eitsupi (changes requested):

R/expr-string.R
etiennebacher: Thanks, fixed

eitsupi (approved):
Thanks!

PR #164, opened 2025-02-27
feat: Add `write_json()` and `write_ndjson()` for DataFrame

etiennebacher:
~I'll add tests once we clear up the testing organization for IO functions (cf https://github.com/eitsupi/neo-r-polars/pull/162#discussion_r1973945477)~


src/rust/src/dataframe/io.rs
```diff
@@ -130,4 +130,27 @@ impl PlRDataFrame {
 
         Ok(out.into())
     }
+
+    #[cfg(not(target_arch = "wasm32"))]
```
etiennebacher, line 134: `json` is listed in `WASM_INCOMPATIBLE_FEATURES` so I suppose this is necessary

etiennebacher:
@eitsupi please let me know if having this test structure is ok for you:

```
test-csv-read-scan.R
test-csv-sink-write.R
test-parquet-read-scan.R
test-parquet-sink-write.R
```

It departs from the fact that sink methods are in lazyframe-frame.R and read methods are in dataframe-frame.R but I think it would be clearer in tests.

I'll finish this PR once this is resolved (no particular rush from my side)

eitsupi:
I think it would be acceptable to move only IO-related functions to separate files, since in practice the current division of R files is considerably distorted by the constraints imposed by the Python class definitions in Python Polars.
And could the division of the test file be changed to accommodate this?

In any case, that would be better done in a separate PR.

tests/testthat/test-io-json-functions.R
```diff
@@ -88,3 +88,40 @@ test_that("scan_ndjson/read_ndjson error", {
...
   expect_snapshot(pl$scan_ndjson("foo", batch_size = 0), error = TRUE)
 })
+
+test_that("write_json: path works", {
+  skip_if_not_installed("jsonlite")
+  dat <- as_polars_df(mtcars[1:5, 1:3])
+  tmpf <- withr::local_tempfile()
+
+  dat$write_json(tmpf)
+  expect_snapshot_file(tmpf, name = "write_json_output")
```
etiennebacher, line 98: Not sure this is actually needed.

PR #165, opened 2025-02-28
feat: Add `get_column()` and `get_column_index()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -244,6 +244,61 @@ dataframe__get_columns <- function() {
...
+#'   error = function(e) print(e)
+#' )
+dataframe__get_column <- function(name, ..., default = NULL) {
+  wrap({
+    tryCatch(
+      self$`_df`$get_column(name),
+      error = function(e) {
+        if (!is.null(default)) {
+          return(default)
+        }
```
eitsupi, line 277: I don't see the value of this option existing.
Wouldn't it be more reasonable for users to run `tryCatch` themselves if they want something like this?

etiennebacher: Not all users know about `tryCatch()`. I don't have a specific usecase for this but the same remark could be done about the py-polars implementation so this should be decided upstream IMO.

eitsupi: Checking the context (pola-rs/polars#17176), this seems to be a Python convention.
Therefore, I do not think it is a good idea to emulate it here.

Given some of the ways it can be done on R, such as the following, I don't think there should be an option to at least allow the user to set the default value (not to mention that if `NULL` is returned, the `%||%` operator can take care of it)

``` r
mtcars$foo
#> NULL

tibble::as_tibble(mtcars)$foo
#> Warning: Unknown or uninitialised column: `foo`.
#> NULL

dplyr::pull(mtcars, "foo")
#> Error in `dplyr::pull()`:
#> ! Can't extract columns that don't exist.
#> ✖ Column `foo` doesn't exist.
```

<sup>Created on 2025-02-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher: Sorry, I don't see the point of removing it. In my experience many R users do not know about `tryCatch()` so if we have a way to simply handle the case where a column doesn't exist, I don't see why we'd remove it. Also the default behavior is to error when the column doesn't exist so I don't see this as a deviation from the R behavior shown in your examples.

> (not to mention that if NULL is returned, the %||% operator can take care of it)

I don't get it, if we remove the `default` arg then this function will just error so `%||%` cannot be used. Or do you propose a mechanism to return `NULL` if the column doesn't exist?

eitsupi: I don't think we need a mechanism to set arbitrary default values; wouldn't it be better to always return a warning and `NULL`, like tibble's `$`?
The current implementation cannot return `NULL`, can it?

etiennebacher: >  The current implementation cannot return NULL, can it?

No it cannot

etiennebacher: > wouldn't it be better to always return a warning and NULL, like tibble's $?

If anything, the examples above show that there's no consistency in R about what to do when a column doesn't exist (NULL, NULL + warning, error). Erroring when a column doesn't exist and providing a default value for users who want it seems ok to me.

eitsupi: I don't think it should return anything of unknown type other than `NULL`.
At least, not being able to return `NULL` is a problem here.

Please remove the `default` argument and change to one of the followings:

- Always error.
- Add something like `raise_if_not_exist` argument (like `expr_meta_output_name`)

eitsupi (changes requested):

eitsupi (approved):
Thanks!

PR #166, opened 2025-02-28
feat: Add `sample()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1888,3 +1888,49 @@ dataframe__write_parquet <- function(
...
+  seed = NULL
+) {
+  wrap({
+    check_dots_empty0(...)
+    if (!is.null(fraction) && !is.null(n)) {
+      abort("cannot specify both `n` and `fraction`")
+    }
+    if (is.null(seed)) {
+      seed <- sample.int(10000, 1)
+    }
```
etiennebacher, line 1918: Unclear to me why the Rust part has `seed: Option<u64>` since we cannot missing `seed` but it is what it is.

R/dataframe-frame.R
```diff
@@ -1888,3 +1888,49 @@ dataframe__write_parquet <- function(
...
+      return(
+        self$`_df`$sample_frac(fraction$`_s`, with_replacement, shuffle, seed) |>
+          wrap()
+      )
+    }
+    if (is.null(n)) {
+      n <- 1
+    }
+    if (!is_polars_series(n)) {
+      n <- as_polars_series(n, "")
```
etiennebacher, line 1932: Not a typo, empty name

R/dataframe-frame.R
eitsupi: My guess is that it is because we want to control the value on the Rust side by the Seed setting on the bindings side?
I think we can confirm this behavior with a unit test.

etiennebacher: > we want to control the value on the Rust side by the Seed setting on the bindings side?

I don't understand what this means. My point was just that if on the binding side we replace NULL/None by a value, then `seed` should just be `u64` and not `Option<u64>` in Rust. But that's not very important, I've added a test that just checks there's no error when we let `seed = NULL`

eitsupi (approved):

PR #168, opened 2025-02-28
feat: Add `unstack()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -1888,3 +1888,103 @@ dataframe__write_parquet <- function(
...
+    how <- arg_match0(how, values = c("vertical", "horizontal"))
+    if (length(columns) > 0) {
+      if (is_polars_selector(columns)) {
+        df <- self$select(columns)
+      } else {
+        df <- self$select(!!!columns)
+      }
+    } else {
+      df <- self
+    }
```
etiennebacher, line 1933: I don't remember if we support selectors in dynamic dots or not. If so, `columns` can be replaced by those

R/dataframe-frame.R
```diff
@@ -1889,3 +1889,103 @@ dataframe__write_parquet <- function(
...
+  step,
+  how = c("vertical", "horizontal"),
+  columns = NULL,
+  fill_values = NULL
+) {
+  wrap({
+    check_dots_empty0(...)
+
+    how <- arg_match0(how, values = c("vertical", "horizontal"))
+    if (length(columns) > 0) {
```
eitsupi, line 1926: Don't you need to check the type before checking the length?

eitsupi (changes requested):

R/dataframe-frame.R
etiennebacher: Why? If `columns` doesn't contain anything we take the whole dataset. I could also use `if (is.null(columns))` but I think that's quite equivalent

eitsupi: Unexpected errors may occur because the type of `columns` has not been verified.

etiennebacher: I agree that should be checked in the if condition but there's no point in checking it before  since we don't know if `columns` contains something.

`columns` can take column names or column selectors, and I just noticed that `parse_into_list_of_expressions()` can take a list mixing those, so I propose we replace `columns` by `...` and we do the type checking before calling `parse_into_list_of_expressions()`. Do you agree?

R/dataframe-frame.R
```diff
@@ -2065,3 +2065,107 @@ dataframe__sample <- function(
...
+    dots <- list2(...)
+    if (length(dots) == 0) {
+      df <- self
+    } else {
+      for (i in seq_along(dots)) {
+        if (!is_string(dots[[i]]) && !is_polars_selector(dots[[i]]) && !is_polars_expr(dots[[i]])) {
+          abort("`...` only accepts column names or column selectors.")
+        }
+      }
+      df <- self$select(!!!dots)
```
etiennebacher, line 2113: `parse_into_list_of_expressions()` works on selectors, but then `!!!dots` would error. I think using `list2()` and checking in the loop that we have column names, selectors, or exprs is fine.

etiennebacher:
> I compared this to Python Polars and there is no guarantee that vectorization by `rep()` will work, so I think a type check of `fill_values` is needed.

Indeed it fails if `fill_values = pl$lit(0)` for instance, but I think the following fix should be enough:
```r
fill_values <- rep(list(fill_values), df$width)
```

eitsupi:
> Indeed it fails if `fill_values = pl$lit(0)` for instance, but I think the following fix should be enough:

If `fill_values` is atomic vector, doesn't that fit?

Since there does not appear to be a function for stretching atomic vectors and others in the same way, it would be better to stretch Series after confirming that the function to be added in #173 can be used to convert them to Seires.

etiennebacher:
> If fill_values is atomic vector, doesn't that fit?

I don't understand what this means.

>  I think we need an additional function for this and would like to add it in the main branch first.

I see that you added the function in main. I'd rather let you wrap up this PR since you already know what to do, but let me know if you're short on time and prefer that I do it.

R/dataframe-frame.R
```diff
@@ -2200,3 +2200,114 @@ dataframe__group_by_dynamic <- function(
...
+  wrap({
+    check_dots_unnamed()
+
+    is_named_list <- is.list(fill_values) &&
+      !is.null(names(fill_values)) &&
+      any(names(fill_values) != "")
+    is_scalar_or_expr <- length(fill_values) == 1 || is_polars_expr(fill_values)
+    if (!is.null(fill_values) && !is_named_list && !is_scalar_or_expr) {
+      abort("`fill_value` must be a scalar or a named list.")
+    }
```
etiennebacher, line 2243: I don't think we have a helper to check if input is a named list or a scalar (and I don't think it's worth doing one)

R/dataframe-frame.R
```diff
@@ -2200,3 +2200,114 @@ dataframe__group_by_dynamic <- function(
...
+
+    # We know whether unstacking will create empty cells, so we extend the
+    # data (and potentially fill missing values) before unstacking.
+    n_fill <- n_cols * n_rows - height
+    if (n_fill > 0) {
+      if (is.null(fill_values)) {
+        fill_values <- NA
+      }
+      if (!is.list(fill_values)) {
+        fill_values <- rep(list(fill_values), df$width)
```
etiennebacher, line 2274: I don't understand how the new `is_convertible_to_polars_series()` would help here (as mentioned in your previous comment)

R/dataframe-frame.R
```diff
@@ -2200,3 +2200,114 @@ dataframe__group_by_dynamic <- function(
...
+        names(fill_values) <- names(df)
```
etiennebacher, line 2275: This is to allow passing columns in `fill_values` in a different order than in `...`, e.g. `unstack("a", "b", step = 2, fill_values = list(b = 1, a = 2))`

etiennebacher (commented):
@eitsupi I've improved how `fill_values` is handled and I've added more tests for it.

eitsupi (approved):
@etiennebacher I rewrote some steps. Could you please check them?

etiennebacher:
Thanks for polishing it, looks good to me

PR #179, opened 2025-03-06
feat: Add `hash_rows()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -2200,3 +2200,37 @@ dataframe__group_by_dynamic <- function(
...
+  wrap({
+    check_number_whole(seed)
+    check_number_whole(seed_1, allow_null = TRUE)
+    check_number_whole(seed_2, allow_null = TRUE)
+    check_number_whole(seed_3, allow_null = TRUE)
+    seed_1 <- seed_1 %||% seed
+    seed_2 <- seed_2 %||% seed
+    seed_3 <- seed_3 %||% seed
+    self$`_df`$hash_rows(seed, seed_1, seed_2, seed_3) |>
+      .savvy_wrap_PlRSeries()
```
etiennebacher, line 2234: Internally, py-polars renames seeds to `k0`, `k1` etc. but I'd rather not do that here because if for some reason there's an error from Rust, the message would be confusing as it would mention `k0`, `k1`, etc.

src/rust/src/dataframe/general.rs
```diff
@@ -368,4 +368,25 @@ impl PlRDataFrame {
             .map_err(RPolarsErr::from)?
             .into())
     }
+
+    pub fn hash_rows(
+        &mut self,
+        seed: NumericScalar,
+        seed_1: NumericScalar,
+        seed_2: NumericScalar,
+        seed_3: NumericScalar,
```
etiennebacher, line 377: Ditto

src/rust/src/dataframe/general.rs
```diff
@@ -368,4 +368,25 @@ impl PlRDataFrame {
...
+    ) -> Result<Sexp> {
```
eitsupi, line 378: Did something like `Result<PlRSeries>` not work?

etiennebacher: I thought I had tried it but actually I could simplify, thanks.

eitsupi (approved):
Thanks!

PR #181, opened 2025-03-08
refactor: Reorganize files for input and output functions

### LINKED ISSUE #178: Reorganize I/O functions
Following https://github.com/eitsupi/neo-r-polars/pull/164#issuecomment-2700854483, here's what I have in mind:

```
input-csv-functions.R (contains read and scan functions)
input-ipc-functions.R
...

output-csv-functions.R (contains write and sink functions)
output-ipc-functions.R
...
```
(and same org for tests)

I'd rather not group input and output functions for a single format in the same file because I prefer having separate test files. There are a ton of arguments in each of those functions so mixing tests of input and output functions would make it messy IMO.

@eitsupi thoughts?

etiennebacher:
This PR just moves code around, it doesn't add functionalities or tests.

Close #178


eitsupi (approved):
Thanks!
(I just removed the rdname tags that were missed.)

PR #184, opened 2025-03-08
test: Set the `Config/testthat/start-first` field to improve performance of parallel test runs

eitsupi:


eitsupi:
In my 2-thread environment (default), I could see a shortening from 1m30s to about 1m20s (I decided to do this because `lazyframe-frame` was always the last one to finish).

PR #186, opened 2025-03-09
feat: Support exporting `nanoarrow_array_stream` via the Arrow C Stream interface

eitsupi:


R/series-s3-nanoarrow.R
```diff
@@ -0,0 +1,51 @@
...
+#' @rdname s3-as_nanoarrow_array_stream
+as_nanoarrow_array_stream.polars_series <- function(x, ..., schema = NULL) {
+  wrap({
+    pl_r_series <- if (!is.null(schema)) {
+      # Sinse polars does not support all Arrow types,
+      # we cannot fully control the output schema
+      # TODO: needs warnings? needs fallback to the default method (use `arrow` to cast)?
+      target_dtype <- nanoarrow::basic_array_stream(list(), schema) |>
+        infer_polars_dtype()
+      x$`_s`$cast(target_dtype$`_dt`, strict = TRUE)
```
eitsupi, line 42: @paleolimbot Sorry to bother you, but do you have any guidelines on what to do when `as_nanoarrow_array_stream()`'s S3 method support for the `schema` argument is halfway through?
In the current implementation, if a type not supported by Polars is specified, as shown below, it is half-heartedly ignored.

```r
as_polars_series(1:3) |>
  nanoarrow::as_nanoarrow_array_stream(
    schema = nanoarrow::na_string() # <- string is required here, but string_view is exported
  )
#> <nanoarrow_array_stream string_view>
#>  $ get_schema:function ()
#>  $ get_next  :function (schema = x$get_schema(), validate = TRUE)
#>  $ release   :function ()
```

Should we fall back to using the `arrow` package in such a case? (I would prefer not to rely on the `arrow` package if possible.)
Or do you think it is worth simply warning?

src/rust/src/series/export.rs
```diff
@@ -359,4 +360,20 @@ impl PlRSeries {
...
     }
+
+    // TODO: add `compat_level` argument
+    pub fn to_arrow_c_stream(&self, stream_ptr: Sexp) -> Result<()> {
+        let stream_ptr = unsafe {
+            ExternalPointerSexp::try_from(stream_ptr)?.cast_mut_unchecked::<ArrowArrayStream>()
+        };
+        let field = self.series.field().to_arrow(CompatLevel::newest());
+        let iter = Box::new(self.series.chunks().clone().into_iter().map(Ok));
+        let stream = export_iterator(iter, field);
```
eitsupi, line 371: Maybe something wrong (pola-rs/polars#21660)?

R/series-s3-nanoarrow.R
eitsupi: Continuing to #188

PR #187, opened 2025-03-09
feat: Add `sink_ndjson()` for LazyFrame

etiennebacher:


R/output-json-functions.R
```diff
@@ -1,6 +1,70 @@
...
+      collapse_joins = collapse_joins,
+      no_optimization = no_optimization
+    )
+
+    lf$sink_json(
+      path = path,
+      maintain_order = maintain_order,
+      storage_options = storage_options,
+      retries = retries
+    )
```
etiennebacher, line 63: It's sink_ndjson() in py-polars but sink_json() in rust-polars

src/rust/src/lazyframe/general.rs
```diff
@@ -1345,4 +1346,21 @@ impl PlRLazyFrame {
...
+    ) -> Result<()> {
+        let options = JsonWriterOptions { maintain_order };
+        let _retries = <Wrap<usize>>::try_from(retries)?.0;
+
+        let cloud_options = Some(CloudOptions::default());
+
+        let ldf = self.ldf.clone();
+        let _ = ldf.sink_json(path, options, cloud_options);
+        Ok(())
+    }
```
etiennebacher, line 1365: This differs quite a bit from rust-polars for now because they're adding all the infrastructure to write partitioned data using `SinkTarget` for instance: https://github.com/pola-rs/polars/blob/4e4768f8b0385138790b1e048fe7617d66f6a4c4/crates/polars-python/src/lazyframe/general.rs#L911

 For now, I'd rather have this simple implementation that can write to a file. We don't use cloud options in other sink functions anyway.

R/output-json-functions.R
```diff
@@ -1,6 +1,68 @@
-# Output (ND)JSON functions: sink_json, sink_ndjson, write_json, write_ndjson
+# Output (ND)JSON functions: sink_ndjson, write_json, write_ndjson
+# (there is only sink_ndjson() in py-polars but there is sink_json() in
+# rust-polars).
```
eitsupi, line 3:
```suggestion
# (there is only sink_ndjson() in Python Polars but there is sink_json() in Rust Polars).
```

R/output-json-functions.R
```diff
@@ -1,6 +1,68 @@
...
 
-# TODO: add sink_json / sink_ndjson
+#' Evaluate the query in streaming mode and write to a NDJSON file
+#'
+#' @description
+#' `r lifecycle::badge("experimental")`
+#'
+#' This allows streaming results that are larger than RAM to be written to disk.
+#'
+#' @inheritParams rlang::check_dots_empty0
```
eitsupi, line 12:
```suggestion
#' @inheritParams rlang::args_dots_empty
```

eitsupi (changes requested):

R/output-json-functions.R
etiennebacher: Done but this is an internal comment so it really doesn't matter much as long as we understand it's in the Rust part.

R/output-json-functions.R
etiennebacher: Done

eitsupi (approved):
Thanks!

PR #190, opened 2025-03-10
docs: Fix formatting and description of `group_by_dynamic()`

etiennebacher:
* Added a space around `*` so that it's not considered markdown formatting (using `\*` doesn't work as it keeps the backslash in docs)
* Added `@description` so that the second half of the description doesn't end up in "Details"


eitsupi (approved):
Good catch, thanks.

PR #191, opened 2025-03-10
feat: Add `str$to_titlecase()`

etiennebacher:
The current `r-polars` implementation relies on the nightly feature.

@eitsupi I think you implemented the features to compile the Rust part? I don't really know what to do here, this PR is just to get the ball rolling.


eitsupi:
Yes, we definitely need to add a "nightly" feature as we do not expect the nightly toolchain to be available at build time.
Please wait a moment as a configure script needs to be written.

PR #194, opened 2025-03-10
feat: Add `rechunk()` and `chunk_lengths()` for Series

etiennebacher:


src/rust/src/series/general.rs
```diff
@@ -132,4 +132,17 @@ impl PlRSeries {
...
         (self.series.n_chunks() as i32).try_into()
     }
+
+    fn chunk_lengths(&self) -> Result<Sexp> {
+        let out = self.series.chunk_lengths().collect::<Vec<usize>>();
+        let mut sexp = OwnedIntegerSexp::new(out.len())?;
+        for (i, out) in out.into_iter().enumerate() {
+            let _ = sexp.set_elt(i, out as i32);
+        }
+        Ok(sexp.into())
```
etiennebacher, line 142: Not sure about this, do we have a better way to convert a `Vec<usize>` to `Sexp`?

src/rust/src/series/general.rs
```diff
@@ -132,4 +132,17 @@ impl PlRSeries {
...
+    }
+
+    pub fn rechunk(&self) -> Result<PlRSeries> {
+        Ok(self.series.rechunk().into())
+    }
```
etiennebacher, line 147: py-polars has an arg `in_place` but it means that this function returns a `Result<Option<Self>>`: https://github.com/pola-rs/polars/blob/052416c8bf54d6063872d5c3413af9e6845acb78/crates/polars-python/src/series/general.rs#L109

AFAIK, `savvy` cannot return an `Option` (not sure about that), which is why I dropped the argument.

src/rust/src/series/general.rs
eitsupi: Since usize may exceed the range of i32, I think we need to implement TryFrom for conversion to an Integer vector.

src/rust/src/series/general.rs
eitsupi: Like 031d87182ed574fffe90dabeeb19ee3c3eaa5ec8?

src/rust/src/series/general.rs
etiennebacher: We could replace `out as i32` by `i32::try_from(out)?`. Would that work for you?

src/rust/src/series/general.rs
etiennebacher: Yes, thanks

src/rust/src/series/general.rs
eitsupi: Good point, something like 48540ea seems to be sufficient.

eitsupi (approved):

PR #195, opened 2025-03-10
feat: Enable `how = "align"` and other `"align_*"` variants in `pl$concat()`

etiennebacher:


R/functions-eager.R
```diff
@@ -47,21 +50,15 @@
...
   ...,
-  how = c(
-    "vertical",
-    "vertical_relaxed",
-    "diagonal",
-    "diagonal_relaxed",
-    "horizontal",
-    "align"
-  ),
+  how = "vertical",
```
etiennebacher, line 61: When there are too many options (here 9 with many characters), it clutters the docs. All options are listed in the description of `how` so I don't think it's necessary to take so much space in the docs with this.

eitsupi: How about list some of them? (Maybe only non-relaxed ones)
Like `tibble::as_tibble()`'s `.name_repair` argument.

https://github.com/eitsupi/neo-r-polars/blob/776e95621042b7f51947da2544db661d9b34cbab/man/s3-as_tibble.Rd#L11

etiennebacher: I don't like that, it would be misleading for the user IMO. If I go to the docs and see a list of accepted values for an argument in the "Usage" section then I assume that's all this argument can take.

R/functions-eager.R
```diff
@@ -100,39 +100,41 @@ pl__concat <- function(
     )
   }
 
-  if (how == "align") {
+  if (startsWith(how, "align")) {
     if (!is_polars_df(first) && !is_polars_lf(first)) {
-      abort(r"("align" strategy is only supported on DataFrames and LazyFrames.)")
+      abort(paste0("\"", how, "\" strategy is only supported on DataFrames and LazyFrames.)"))
```
eitsupi, line 105:
```suggestion
      abort(sprintf("'%s' strategy is only supported on DataFrames and LazyFrames.", how))
```

...And please add snapshot tests for this error.

eitsupi (changes requested):

R/functions-eager.R
eitsupi: It is quite a pain for me to have to look at the help page to check the available values, but well, it is checked in error, so this is still not too bad.

PR #205, opened 2025-03-18
feat: `SQLContext` related functions

etiennebacher:
This implements `$sql()` and the infrastructure for it. Not all [SQLContext functions](https://docs.pola.rs/api/python/dev/reference/sql/python_api.html#methods) are supported yet so that this PR is not too big.


R/lazyframe-frame.R
```diff
@@ -2560,4 +2560,49 @@ lazyframe__describe <- function(
...
+#'       SELECT
+#'          a,
+#'          (a % 2 == 0) AS a_is_even,
+#'          (b::float4 / 2) AS 'b/2',
+#'          CONCAT_WS(':', c, c, c) AS c_c_c
+#'       FROM frame
+#'       ORDER BY a
+#' ",
+#' table_name="frame",
+#' )$filter(!pl$col("c_c_c")$str$starts_with("x"))$collect()
```
eitsupi, line 2598: Seems not formatted.

R/sql-context.R
```diff
@@ -0,0 +1,102 @@
...
+}
+
+pl__SQLContext <- function(..., register_globals = FALSE) {
+  wrap({
+    self <- PlRSQLContext$new()
+    frames <- list2(...)
+    for (i in seq_along(frames)) {
+      frame <- ensure_lazyframe(frames[[i]])
+      self$register(names(frames)[i], frame$`_ldf`)
+    }
```
eitsupi, line 26: Needs `register_many()`

eitsupi (changes requested):

R/sql-context.R
etiennebacher: Done.

R/lazyframe-frame.R
etiennebacher: The `=` were indeed not formatted, but `air` doesn't change anything in the query itself.

R/sql-context.R
```diff
@@ -0,0 +1,101 @@
...
+    self <- PlRSQLContext$new() |>
+      wrap()
+    self$register_many(...)
+    self
+  })
+}
+
+ensure_lazyframe <- function(obj) {
+  as_polars_lf(obj)
+}
```
eitsupi, line 30: What is this function needed for?

etiennebacher: It's basically an alias for as_polars_lf(), I thought you'd like to have it for symmetry with py-polars: https://github.com/pola-rs/polars/blob/d95e343e04d47bcbeafc815a2d9855aa5099f1bb/py-polars/polars/sql/context.py#L471

But we can drop it ofc.

etiennebacher: After the change due to earlier comment, it's basically an alias for as_polars_lf(), I thought you'd like to have it for symmetry with py-polars: https://github.com/pola-rs/polars/blob/d95e343e04d47bcbeafc815a2d9855aa5099f1bb/py-polars/polars/sql/context.py#L471

But we can drop it ofc.

R/sql-context.R
```diff
@@ -44,6 +38,9 @@ sql_context__register <- function(name, frame = NULL) {
 sql_context__register_many <- function(...) {
   wrap({
     frames <- list2(...)
+    if (is.null(names(frames)) || any(names(frames) == "")) {
```
eitsupi, line 41: It seems that `is_named2()` can be used here.

eitsupi (changes requested):

R/sql-context.R
eitsupi: I think this is only necessary because Python does not have a generic function for such a use, and it is not needed here.

etiennebacher: Done

R/sql-context.R
etiennebacher: Done

R/sql-context.R
```diff
@@ -0,0 +1,190 @@
...
+#' @export
+wrap.PlRSQLContext <- function(x) {
+  self <- new.env(parent = emptyenv())
+  self$`_ctxt` <- x
+
+  class(self) <- c("polars_sql_context", "polars_object")
+  self
+}
+
+get_frame_locals <- function(all_compatible) {
```
etiennebacher, line 13: Currently this only looks in the global env and not in all parent frames as in py-polars: https://github.com/pola-rs/polars/blob/main/py-polars/polars/_utils/various.py#L499

(`_get_stack_locals()` is called from `_get_frame_locals()`)

I found it surprisingly difficult to collect all objects in all parent frames without running into recursion issues.

tests/testthat/test-sql-context.R
```diff
@@ -0,0 +1,27 @@
...
+  expect_equal(
+    pl$SQLContext(mtcars = mtcars)$tables(),
+    "mtcars"
+  )
+  expect_equal(
+    pl$SQLContext(mtcars = mtcars, .register_globals = TRUE)$tables(),
+    c("foo", "mtcars")
+  )
+  rm("foo", envir = global_env())
+})
```
etiennebacher: Not super clean but we need `foo` to be in the global environment (cf previous comment). AFAIK `withr` doesn't allow this and `assign()` is better than `<<-`.

tests/testthat/test-sql-context.R
```diff
@@ -0,0 +1,37 @@
...
+  )
+  # TODO: add message pattern when this is fixed
+  # https://github.com/eitsupi/neo-r-polars/issues/206
+  expect_error(pl$SQLContext(a = complex(1)))
+})
+
+test_that("arg '.register_globals' works", {
+  foo <- pl$DataFrame(x = 1)
+  assign("foo", foo, envir = global_env())
+  expect_equal(
```
etiennebacher, line 18: Not super clean but we need `foo` to be in the global environment (cf previous comment). AFAIK `withr` doesn't allow this and `assign()` is better than `<<-`.

eitsupi (changes requested):
Could you not add `.register_globals` for now?
This is quite complex and should be added later. (In my opinion we need to be able to scan arbitrary environments by passing the environment as an argument)

eitsupi:
In other words, there is a risk of tremendous performance degradation due to the need to inspect every object in the environment.

etiennebacher:
> Could you not add `.register_globals` for now?

Sure, happy to remove some complexity.

R/sql-context.R
```diff
@@ -0,0 +1,157 @@
...
+#' df <- pl$DataFrame(x = 1)
+#' ctx <- pl$SQLContext()
+#' ctx$register("frame_data", df)$execute("SELECT * FROM frame_data")$collect()
+sql_context__register <- function(name, frame = NULL) {
+  wrap({
+    frame <- if (!is.null(frame)) {
+      as_polars_lf(frame)
+    } else {
+      pl$LazyFrame()
+    }
```
eitsupi, line 56: I think this case separation is unnecessary since `as_polars_lf(NULL)` works.

R/sql-context.R
```diff
@@ -0,0 +1,157 @@
...
+#' df3 <- pl$DataFrame(x = 3)
+#'
+#' ctx <- pl$SQLContext()
+#' ctx$register_many(tab1 = df, tab2 = df2, tab3 = df3)
+sql_context__register_many <- function(...) {
+  wrap({
+    frames <- list2(...)
+    if (length(frames) == 0) {
+      return(self)
+    }
```
eitsupi, line 77: Is this necessary?

eitsupi (changes requested):

R/sql-context.R
etiennebacher: Ah nice, thanks.

R/sql-context.R
etiennebacher: Not anymore, it was at some point. Fixed.

eitsupi (approved):
Looks good, thanks!

PR #207, opened 2025-03-22
feat: Better error messages for the default methods of `as_polars_df()`, `as_polars_series()` and `infer_polars_dtype()`

### LINKED ISSUE #206: Improve error message for `as_polars_lf()` conversion failures
Currently, when conversion to LazyFrame fails, the error message doesn't mention `as_polars_lf()`, which makes it harder to track where the error comes from:

```r
as_polars_lf(complex(1))

Error:
! Unsupported class for `infer_polars_dtype()`: complex
Caused by error in `infer_polars_dtype_default_impl()`:
! Unsupported class for `as_polars_series()`: complex
```

eitsupi:
Close #206

### DIFF
--- a/R/as_polars_df.R
+++ b/R/as_polars_df.R
@@ -78,7 +78,10 @@ as_polars_df.default <- function(x, ...) {
     infer_polars_dtype(x, ...),
     error = function(cnd) {
       abort(
-        "This object is not supported for the default method of `as_polars_df()` because it can't be converted to a polars Series.",
+        sprintf(
+          "%s may not be converted to a polars Series, and hence to a polars DataFrame.",
+          obj_type_friendly(x)
+        ),
         parent = cnd
       )
     }
--- a/R/as_polars_df.R
+++ b/R/as_polars_df.R
@@ -74,7 +74,15 @@ as_polars_df <- function(x, ...) {
 #' @rdname as_polars_df
 #' @export
 as_polars_df.default <- function(x, ...) {
-  dtype <- infer_polars_dtype(x, ...)
+  dtype <- try_fetch(
+    infer_polars_dtype(x, ...),
+    error = function(cnd) {
+      abort(
+        "This object is not supported for the default method of `as_polars_df()` because it can't be converted to a polars Series.",
+        parent = cnd
+      )
+    }
+  )
   if (inherits(dtype, "polars_dtype_struct")) {
     as_polars_series(x, ...)$struct$unnest()
   } else {
@@ -83,8 +91,7 @@ as_polars_df.default <- function(x, ...) {
       c(
         "This object is not supported for the default method of `as_polars_df()` because it is not a Struct dtype like object.",
         i = "Use `infer_polars_dtype()` to check the dtype for corresponding to the object."
-      ),
-      call = parent.frame()
+      )
     )
   }
 }
--- a/R/as_polars_lf.R
+++ b/R/as_polars_lf.R
@@ -19,7 +19,15 @@ as_polars_lf <- function(x, ...) {
 #' @rdname as_polars_lf
 #' @export
 as_polars_lf.default <- function(x, ...) {
-  as_polars_df(x, ...)$lazy()
+  try_fetch(
+    as_polars_df(x, ...),
+    error = function(cnd) {
+      abort(
+        "Failed to create a polars DataFrame.",
+        parent = cnd
+      )
+    }
+  )$lazy()
 }
 
 #' @rdname as_polars_lf
--- a/R/as_polars_lf.R
+++ b/R/as_polars_lf.R
@@ -20,14 +20,14 @@ as_polars_lf <- function(x, ...) {
 #' @export
 as_polars_lf.default <- function(x, ...) {
   try_fetch(
-    as_polars_df(x, ...),
+    as_polars_df(x, ...)$lazy(),
     error = function(cnd) {
       abort(
-        "Failed to create a polars DataFrame.",
+        "Failed to create a polars LazyFrame.",
         parent = cnd
       )
     }
-  )$lazy()
+  )
 }
 
 #' @rdname as_polars_lf
--- a/R/as_polars_series.R
+++ b/R/as_polars_series.R
@@ -182,7 +182,7 @@ as_polars_series <- function(x, name = NULL, ...) {
 #' @rdname as_polars_series
 #' @export
 as_polars_series.default <- function(x, name = NULL, ...) {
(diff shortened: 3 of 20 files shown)


R/as_polars_df.R
```diff
@@ -74,7 +74,15 @@ as_polars_df <- function(x, ...) {
 #' @rdname as_polars_df
 #' @export
 as_polars_df.default <- function(x, ...) {
-  dtype <- infer_polars_dtype(x, ...)
+  dtype <- try_fetch(
+    infer_polars_dtype(x, ...),
+    error = function(cnd) {
+      abort(
+        "This object is not supported for the default method of `as_polars_df()` because it can't be converted to a polars Series.",
```
etiennebacher, line 81: This could be made shorter IMO, e.g:

```suggestion
       paste0("Cannot convert ", obj_type_friendly(x), " to a Polars Series, and hence to a Polars DataFrame.")
```
or sth similar.

R/as_polars_series.R
```diff
@@ -182,10 +182,7 @@ as_polars_series <- function(x, name = NULL, ...) {
 #' @rdname as_polars_series
 #' @export
 as_polars_series.default <- function(x, name = NULL, ...) {
-  abort(
-    paste0("Unsupported class for `as_polars_series()`: ", toString(class(x))),
-    call = parent.frame()
-  )
+  abort(paste0("Unsupported class for `as_polars_series()`: ", toString(class(x))))
```
etiennebacher, line 185: Can we use `obj_type_friendly()`?

tests/testthat/_snaps/as_polars_lf.md
```diff
@@ -0,0 +1,21 @@
+# as_polars_lf.default throws an error
+
+    Code
+      as_polars_lf(1)
+    Condition
+      Error in `as_polars_df()`:
```
etiennebacher, line 6: It is confusing that the call is `as_polars_lf()` but the message points to `as_polars_df()`

etiennebacher (changes requested):
Thanks, looks better but I think there are still improvements to be made.

tests/testthat/_snaps/as_polars_lf.md
eitsupi: I see what you mean, but since `as_polars_lf.default()` is a thin wrapper around `as_polars_df()`, as the documentation states, I'm wondering if it's worth doing another `tryCatch` for the error message.

etiennebacher: I think it's worth it since it only needs to be implemented in `as_polars_lf.default`. It would be different if we needed an extra tryCatch() in several methods.

R/as_polars_series.R
eitsupi: Good point. I did not know that I could use that function because I had been using `toString()` since before I added the standalone file.

R/as_polars_df.R
```diff
@@ -74,7 +74,18 @@ as_polars_df <- function(x, ...) {
 #' @rdname as_polars_df
 #' @export
 as_polars_df.default <- function(x, ...) {
-  dtype <- infer_polars_dtype(x, ...)
+  dtype <- try_fetch(
+    infer_polars_dtype(x, ...),
+    error = function(cnd) {
+      abort(
+        sprintf(
+          "%s may not be converted to a polars Series, and hence to a polars DataFrame.",
```
etiennebacher, line 82:
```suggestion
          "%s can't be converted to a polars Series, and hence to a polars DataFrame.",
```

R/as_polars_lf.R
```diff
@@ -19,7 +19,15 @@ as_polars_lf <- function(x, ...) {
 #' @rdname as_polars_lf
 #' @export
 as_polars_lf.default <- function(x, ...) {
-  as_polars_df(x, ...)$lazy()
+  try_fetch(
+    as_polars_df(x, ...),
+    error = function(cnd) {
+      abort(
+        "Failed to create a polars DataFrame.",
```
etiennebacher, line 26:
```suggestion
        "Failed to create a polars LazyFrame.",
```
Also why not put the `$lazy()` in `try_fetch()`?

R/as_polars_series.R
```diff
@@ -182,10 +182,7 @@ as_polars_series <- function(x, name = NULL, ...) {
 #' @rdname as_polars_series
 #' @export
 as_polars_series.default <- function(x, name = NULL, ...) {
-  abort(
-    paste0("Unsupported class for `as_polars_series()`: ", toString(class(x))),
-    call = parent.frame()
-  )
+  abort(sprintf("Unsupported object can be converted to a polars Series: %s", obj_type_friendly(x)))
```
etiennebacher, line 185:
```suggestion
  abort(sprintf("%s can't be converted to a polars Series", obj_type_friendly(x)))
```

etiennebacher (changes requested):

R/as_polars_lf.R
eitsupi: I would like to isolate here whether the creating a DataFrame failed to create or failed in lazy.

R/as_polars_df.R
eitsupi: Strictly speaking, the failure of `infer_polars_dtype()` does not guarantee that `as_polars_series()` will fail.

R/as_polars_lf.R
etiennebacher: Yeah but then we're back to the same issue where the user calls `as_polars_lf()` and sees `"Failed to create a polars DataFrame"`

R/as_polars_df.R
etiennebacher: It doesn't? Do we have examples of this?

eitsupi: For example, nanoarrow_array_stream.
I have added `as_polars_series` and `infer_polars_dtype` methods at the same time so `as_polars_df(<nanoarrow_array_stream>)` works, but if I register only `as_polars_series.nanoarrow_array_stream`, `infer_polars_dtype(<nanoarrow_array_stream>)` and `as_polars_df(<nanoarrow_array_tream>)` will fail.

R/as_polars_lf.R
etiennebacher: Also, I think the `parent = cnd` will show the error originating from `as_polars_df()`. So the user will still see in the trace that the source of the error comes from the creation of the polars DataFrame, but the main message should be "Failed to create a polars LazyFrame".

etiennebacher (approved):
LGTM, thanks!

PR #209, opened 2025-03-22
ci: Enable sccache on wasm build test on ci

eitsupi:
Ref: <https://depot.dev/blog/sccache-in-github-actions>


eitsupi:
Unfortunately, this does not seem to work. The build time did not decrease between the first and second builds.

PR #217, opened 2025-03-23
ci: Simplify jobs for faster testing

### LINKED ISSUE #216: Do we need R CMD checks on all platforms?
R CMD checks seem to require a significant amount of time in areas other than simple test-like things.
https://github.com/eitsupi/neo-r-polars/blob/07e44a38e85923b6372bb6c65a1d289df907b466/.github/workflows/test-r.yml#L106-L109

For example, if we were to do only simple tests using debug builds on all platforms and R versions, and `R CMD check` only on r-devel on Ubuntu, the CI iterations would be far faster.

eitsupi:
Close #216

### DIFF
--- a/.github/workflows/test-r.yml
+++ b/.github/workflows/test-r.yml
@@ -166,12 +166,6 @@ jobs:
           # we use the debug build here.
           Rscript -e 'pkgbuild::compile_dll(debug = TRUE)'
 
-      - name: Install package
-        env:
-          DEBUG: "true"
-        run: |
-          R CMD INSTALL --no-multiarch --with-keep.source .
-
       - name: Run tests
         shell: Rscript {0}
         run: |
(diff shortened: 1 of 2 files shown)


eitsupi:
I have disabled tests on Windows for now because I get the following error on Windows and cannot run the tests.
It is unclear why the error occurs.

```log
 Error in `private$handle_error()`:
! testthat subprocess failed to start, stderr:
Error in inDL(x, as.logical(local), as.logical(now), ...) : 

  unable to load shared object 'C:/Users/RUNNER~1/AppData/Local/Temp/RtmpSIRl9D/pkgload1aec417b5f70/neopolars.dll':

  LoadLibrary failure:  %1 is not a valid Win32 application.

Calls: callr_startup_hook ... <Anonymous> -> load_dll -> library.dynam2 -> dyn.load -> inDL
```

PR #229, opened 2025-03-23
feat: Add `glimpse()` for DataFrame

etiennebacher:


R/dataframe-frame.R
```diff
@@ -2216,3 +2216,90 @@ dataframe__describe <- function(
...
+#' df$glimpse()
+#'
+#' df$glimpse(max_items_per_column = 3)
+#'
+#' df$glimpse(max_items_per_column = 3, max_colname_length = 3)
+dataframe__glimpse <- function(
+  ...,
+  max_items_per_column = 10,
+  max_colname_length = 50
+) {
```
etiennebacher, line 2244: py-polars has an arg `return_as_string` but as in other functions we don't necessarily want this kind of arg. Here I always invisibly return the string.

R/dataframe-frame.R
```diff
@@ -2216,3 +2216,90 @@ dataframe__describe <- function(
...
+        dat$dtype_str <- paste0(" <", dat$dtype_str, ">")
+      }
+      output <- paste0(
+        output,
+        paste0("$ ", dat$col_name, dat$dtype_str, ": ", dat$val_str, "\n")
+      )
+    }
+    cat(output)
+  })
+  invisible(output)
```
etiennebacher, line 2304: I put this outside `wrap()` otherwise it prints the string in addition to `cat()`

src/rust/src/functions/misc.rs
eitsupi, line 1: Do not place this here, make it a method of dtype.

Perhaps just adding an argument to the existing `as_str` to switch between the two representation methods would be sufficient.

eitsupi (changes requested):

src/rust/src/functions/misc.rs
etiennebacher: This function and this file exist in rust-polars: https://github.com/pola-rs/polars/blob/0aa889bef845ed902c634f900d4ed703c2eb55aa/crates/polars-python/src/functions/misc.rs#L10

eitsupi: Thanks for pointing that out. That makes sense.

Even in that case, given that polars-python does not have a dtype with methods on the Rust side unlike r-polars, I think it is better to make it a method in r-polars (see how the `max` method of the dtype on Python Polars implemented).

etiennebacher: I've removed the function and added `abbreviated` in `as_str()` for datatypes.

R/datatypes-s3-base.R
```diff
@@ -38,7 +38,7 @@
 
 #' @export
 print.polars_dtype <- function(x, ...) {
-  x$`_dt`$as_str() |>
+  x$as_str() |>
```
etiennebacher, line 41: Necessary now that `as_str()` is properly added in datatypes functions on the R side. (The alternative would be to manually add `abbreviated = FALSE`, which isn't better)

R/datatypes-s3-waldo.R
```diff
@@ -1,7 +1,7 @@
 # exported in zzz.R
 compare_proxy.polars_dtype <- function(x, path) {
   list(
-    object = x$`_dt`$as_str(),
+    object = x$as_str(),
```
etiennebacher, line 4: Ditto

R/datatypes-classes.R
```diff
@@ -280,3 +280,11 @@ datatype__min <- function() {
...
     wrap()
 }
+
+#' @param abbreviated Use short form of datatype name, used for instance in
+#' [`glimpse()`][dataframe__glimpse].
+#' @noRd
+datatype__as_str <- function(abbreviated = FALSE) {
+  self$`_dt`$as_str(abbreviated = abbreviated) |>
+    wrap()
+}
```
eitsupi, line 290: Why do you need to add this?
(I do not want to add this. I was thinking that perhaps something like the S3 method of the format function needs to be added instead.)

eitsupi (changes requested):

R/datatypes-classes.R
etiennebacher: I don't understand the issue, it's simply here so that we can do `x$as_str()` instead of `x$`_dt`$as_str()`. Why is this a problem? I also don't see any S3 methods for `format()` currently so I don't think it should be added in this PR.

eitsupi: In other words, I am not sure if the method `as_str` of dtype should be made public. (That's why it hasn't been added so far)

etiennebacher: So you'd rather have `format.polars_dtype`?

eitsupi: Yes.
So just made that a TODO and now can't use `` x$`_dt`$as_str() ``?

R/dataframe-frame.R
```diff
@@ -2256,3 +2256,90 @@ dataframe__count <- function() {
...
+) {
+  wrap({
+    check_dots_empty0(...)
+    max_n_values <- min(max_items_per_column, self$height)
+    schema <- self$schema
+
+    parse_column <- \(col_name, dtype) {
+      values <- self$select(pl$col(col_name)$slice(0, max_n_values)) |>
+        as.list()
+      val_str <- toString(values[[1]])
```
eitsupi, line 2293: (I have not checked the Python Polars implementation.)

Unexpected results may occur because in the future it may be possible to change the default values of arguments in the `as.list(<DataFrame>)` with the options function, similar to the current r-polars.
Wouldn't it be better to use cast in Polars to make it into a string and then return it to the R side?

etiennebacher: Ah yes, that's a good point. I've added a cast to String which slightly changes the output (e.g "5" is replaced by "5.0") but I think it's fine.

I've also added an int64 column in the snapshots to catch this kind of changes in `as.list()`.

etiennebacher: I keep the int8 column on purpose in snapshots to check the printing of abbreviated data types that don't have the same length.

eitsupi (approved):
Thanks!

PR #247, opened 2025-03-29
feat: Add `select_seq()` for DataFrame

etiennebacher:


R/lazyframe-frame.R
```diff
@@ -110,7 +110,7 @@ lazyframe__select <- function(...) {
 #'   bar = 6:8,
 #'   ham = letters[1:3]
 #' )
-#' lf$select_seq("foo")$collect()
+#' lf$select_seq("foo", bar2 = pl$col("bar") * 2)$collect()
```
etiennebacher, line 113: Small drive-by: just show that this function is not only for selection but can also create new vars

eitsupi (approved):
Thanks!

PR #258, opened 2025-04-02
chore: Update standalone `rlang` files

etiennebacher:
Those were recently reformatted in `rlang` so nothing to be done on our side.


eitsupi:
I'm confused by the title.
You just updated it to the latest files, right?

It may be a detailed account of what happened upstream, but I can't tell from your comment that these files were updated correctly.

eitsupi (approved):
I see, thanks!

PR #264, opened 2025-04-03
chore: Deprecate the `streaming` argument and document the `engine` argument

eitsupi:


R/lazyframe-frame.R
```diff
@@ -230,6 +235,9 @@ lazyframe__collect <- function(
     engine <- arg_match0(engine, c("auto", "in-memory", "streaming", "old-streaming"))
     # TODO: remove the streaming argument
     if (isTRUE(streaming)) {
+      deprecate_warn(
+        "The `streaming` argument is deprecated and will be removed in future. Use `engine = 'old-streaming'` instead."
+      )
```
etiennebacher, line 240: Should this mention the new streaming engine? Maybe like this:

> The `streaming` argument is deprecated and will be removed in the future. Use `engine = 'old-streaming'` for identical behavior, and `engine = 'streaming'` to use the new streaming engine.

The average user doesn't necessarily know about the transition from the old to the new engine, so it's better to make them aware of this IMO.

eitsupi: Thanks, that makes sense.
I'll update that and add tests.

PR #275, opened 2025-04-09
fix: Ensure install the target for nightly toolchain by the configure script (especially for windows gnu)

eitsupi:


configure
```diff
@@ -181,6 +177,18 @@ So using the rust-toolchain.toml file to set the toolchain.
 --------------------------------------------------------------------
 EOF
     copy_rust_toolchain_toml
+
+    # Try to add the target
+    # TODO: Tier 2 without Host Tools support
```
eitsupi, line 182:
```suggestion
```

PR #284, opened 2025-04-10
docs(news): Trim and reorganize NEWS

etiennebacher:


eitsupi:
I don't know if it's worth having as a stand-alone article about the differences from previous versions of the package; is there any reason why it shouldn't be the NEWS file?

eitsupi:
Thanks, I think that's a good direction.

It might be worth mentioning that it is possible to pass Series and DataFrame between polars0 and neopolars via nanoarrow array stream as shown below (Indicates that PTime can be converted to hms)

``` r
ptime_vec <- polars0::as_polars_series(seq(1e6, 5e6, length.out = 5))$cast(polars0::pl$Time)$to_r()
ptime_vec
#> PTime [ double ]: number of epochs [ ns ] since midnight
#> [1] "00:00:00:001_000_000ns val: 1e+06" "00:00:00:002_000_000ns val: 2e+06"
#> [3] "00:00:00:003_000_000ns val: 3e+06" "00:00:00:004_000_000ns val: 4e+06"
#> [5] "00:00:00:005_000_000ns val: 5e+06"

series <- polars0::as_polars_series(ptime_vec) |>
  nanoarrow::as_nanoarrow_array_stream() |>
  neopolars::as_polars_series()
#> Registered S3 methods overwritten by 'neopolars':
#>   method               from
#>   print.polars_info    polars0
#>   print.polars_options polars0

series$to_r_vector()
#> 00:00:00.001
#> 00:00:00.002
#> 00:00:00.003
#> 00:00:00.004
#> 00:00:00.005
```

<sup>Created on 2025-06-08 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher:
> It might be worth mentioning that it is possible to pass Series and DataFrame between polars0 and neopolars via nanoarrow array stream

This means someone would use both the old and new versions of `polars` in the same R session, which I think we should discourage since the old one will not be maintained at all. The NEWS for 1.0.0 will be quite long so I'd rather let this item aside to avoid making it even more complex.

NEWS_abbr.md
```diff
@@ -0,0 +1,288 @@
...
+  series <- pl$Series("foo", 1)
+  pl$DataFrame(bar = 1:2)$with_columns(series)
+  ```
+
+* `map_batches()` and `map_elements()` still exist but their usage is
+  discouraged. Those functions are not guaranteed to interact correctly with
+  the streaming engine. To apply functions from external packages or custom
+  functions that cannot be translated to polars syntax, we now recommend
+  converting the data to a `data.frame` and use the parallel framework of
+  `purrr` for efficiency. The vignette "Using custom functions" contains more
```
etiennebacher, line 284: This is the current state in the rewrite as I understand it, but maybe I missed something.

Btw, if this paragraph is correct, then I guess we should also keep `map_elements()`? Currently only `map_batches()` is implemented.

NEWS_abbr.md
```diff
@@ -0,0 +1,288 @@
...
+  details about this.
```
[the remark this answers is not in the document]
eitsupi: I looked at the implementation of `map_elements` and could not understand the significance of its existence.\
I think it would be sufficient to just return the Series to a vector once and execute a for loop, etc.
In other words, it is not worth implementing.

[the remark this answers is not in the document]
etiennebacher: I think it would mostly be a convenience function to directly pass scalar inputs to a function, so that we don't have to write a loop ourselves:

```r
pl$col("foo")$map_elements(my_fun_that_takes_scalar)
```
is just nicer to use than:
```r
pl$col("foo")$map_batches(\(x) {
  elems <- as.vector(x)
  vapply(elems, my_fun_that_takes_scalar, FUN.VALUE = sth)
})
```

However, I hadn't realized their implementation were so different, I thought we could just reuse most of the code of `map_batches()`. I'm fine with leaving it aside since we have an entire vignette about using `purrr` anyway.

etiennebacher:
@eitsupi I think this is ready for review. I can move the changes from NEWS_abbr.md to NEWS.md once the PR is approved (I think it's just easier to review in a separate file to not have the git diff).

NEWS_abbr.md
```diff
@@ -0,0 +1,287 @@
...
+  #> Run `rlang::last_trace()` to see where the error occurred.
+  ```
+
+* Conversion from `polars` to R types has been revamped: `$to_r()`, `$to_list()`
+  and `$to_data_frame()` no longer exist. Instead, you must use `as.data.frame()`,
+  `as.list()` and `as.vector()`.
+
+  If finer control is needed (for instance to specify how integer-64 or date
+  columns should be handled), use either `$to_r_vector()` or set options for the
+  entire session with `polars_options()`.
```
eitsupi, line 64: I think the description of the difference between `as.vector()` and `$to_r_vector()` is inaccurate.
For example, it should not be possible to convert to integer64 with `as.vector()` (because the attribute is removed)

NEWS_abbr.md
```diff
@@ -0,0 +1,287 @@
...
+
+* In general, `polars` now uses dots (`...`) in two scenarios:
+
+  1. to pass an unlimited number of inputs (for instance in `select()`, `cast()`,
+     or `group_by()`), using [dynamic-dots](https://rlang.r-lib.org/reference/dyn-dots.html).
+
+    For example, if you used to pass a vector of column names or a list of
+    expressions, you now need to expand it with `!!!`:
```
eitsupi, line 72: (At least on GitHub) listings are not considered contiguous.
Please adjust for blank lines and line breaks.

NEWS_abbr.md
```diff
@@ -0,0 +1,287 @@
...
+    #> ┌──────┬──────┬─────┐
+    #> │ drat ┆ mpg  ┆ cyl │
+    #> │ ---  ┆ ---  ┆ --- │
+    #> │ f64  ┆ f64  ┆ f64 │
+    #> ╞══════╪══════╪═════╡
+    #> │ 4.9  ┆ 21.0 ┆ 6.0 │
+    #> │ 4.9  ┆ 21.0 ┆ 6.0 │
+    #> │ 4.85 ┆ 22.8 ┆ 4.0 │
+    #> └──────┴──────┴─────┘
+    ```
```
eitsupi, line 101: Could you split code blocks old and new?

eitsupi (changes requested):

NEWS_abbr.md
etiennebacher: [f5b22e8](https://github.com/eitsupi/neo-r-polars/pull/284/commits/f5b22e881fcd10609a381ef5e4c40f5fa5dff717) should be more accurate like this.

NEWS_abbr.md
etiennebacher: Indeed, this part was missing a white space in indentation

NEWS_abbr.md
etiennebacher: [5ea8298](https://github.com/eitsupi/neo-r-polars/pull/284/commits/5ea829822d6ab2d818e403bb17a582a313a09446)

NEWS_abbr.md
```diff
@@ -0,0 +1,304 @@
...
+  #>
+  #>     If you want expression: Series[foo] to be broadcasted, ensure it is a
+  #>     scalar (for instance by adding '.first()').
+  ```
+
+  ```r
+  ### NEW
+  series <- pl$Series("foo", 1)
+  pl$DataFrame(bar = 1:2)$with_columns(series)
+  ```
```
etiennebacher, line 294: This needs #432 to be resolved

NEWS_abbr.md
```diff
@@ -0,0 +1,287 @@
...
+
+
+### Other changes
+
+* Passing a Series with a single value in `$with_columns()` now works:
+
+  ```r
+  series <- pl$Series("foo", 1)
+  pl$DataFrame(bar = 1:2)$with_columns(series)
+  ```
```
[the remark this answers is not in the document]
etiennebacher: Now looks like this:

* R objects that convert to a Series of length 1 are now treated like scalar
  values when converting to `polars` Expr:

  ```r
  ### OLD
  series <- pl$Series("foo", 1)
  pl$DataFrame(bar = 1:2)$with_columns(series)
  #> [...truncated...]
  #> Encountered the following error in Rust-Polars:
  #>     	Series foo, length 1 doesn't match the DataFrame height of 2
  #>
  #>     If you want expression: Series[foo] to be broadcasted, ensure it is a
  #>     scalar (for instance by adding '.first()').
  ```

  ```r
  ### NEW
  series <- pl$Series("foo", 1)
  pl$DataFrame(bar = 1:2)$with_columns(series)
  #> shape: (2, 2)
  #> ┌─────┬─────┐
  #> │ bar ┆ foo │
  #> │ --- ┆ --- │
  #> │ i32 ┆ f64 │
  #> ╞═════╪═════╡
  #> │ 1   ┆ 1.0 │
  #> │ 2   ┆ 1.0 │
  #> └─────┴─────┘
  ```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
 ```
 
-```r
-# New version
-pl$col(!!!c("foo", "bar"), "baz")
-#> cols(["foo", "bar", "baz"])
-```
+### Breaking changes
+
+* The class names of `polars` objects have changed:
```
eitsupi, line 26:
```suggestion
* The class names of polars objects have changed:
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  - `RPolarsLazyFrame` -> `polars_lazy_frame`
+  - `RPolarsDataFrame` -> `polars_data_frame`
+  - `RPolarsSeries` -> `polars_series`
+  - `RPolarsExpr` -> `polars_expr`
+
+* Conversion from unknown classes to Polars objects now fails. Developers can
+  specify how those objects should be handled by `polars` by creating a method
```
eitsupi, line 33:
```suggestion
  specify how those objects should be handled by polars by creating a method
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  for `as_polars_series.<my-object>`.
```
eitsupi, line 34:
```suggestion
  for `as_polars_series.my_class`.
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  ### NEW
+  a <- 1
+  class(a) <- "foo"
+  as_polars_series(a)
+  #> Error:
+  #> a <foo> object can't be converted to a polars Series.
+  #> Run `rlang::last_trace()` to see where the error occurred.
+  ```
+
+* Conversion from `polars` to R types has been revamped: `$to_r()`, `$to_list()`
```
eitsupi, line 58:
```suggestion
* Conversion from polars objects to R vectors has been revamped: `$to_r()`, `$to_list()`
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  and `$to_data_frame()` no longer exist. Instead, you must use `as.data.frame()`,
+  `as.list()` and `as.vector()`.
+
+  `as.vector()` will remove attributes that might be useful, for instance to
+  convert Int64 values using the `bit64` package or to convert Time values using
```
eitsupi, line 63:
```suggestion
  convert Int64 values using the bit64 package or to convert Time values using
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  the `hms` package. If finer control is needed, use either `$to_r_vector()` or
```
eitsupi, line 64:
```suggestion
  the hms package. If finer control is needed, use either `$to_r_vector()` or
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  set options for the entire session with `polars_options()`.
+
+* In general, `polars` now uses dots (`...`) in two scenarios:
```
eitsupi, line 67:
```suggestion
* In general, polars now uses dots (`...`) in two scenarios:
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  #> │ x   ┆ y   │
+  #> │ --- ┆ --- │
+  #> │ f64 ┆ str │
+  #> ╞═════╪═════╡
+  #> │ 1.0 ┆ a   │
+  #> └─────┴─────┘
+  ```
+
+  Use `as_polars_df()` and `as_polars_lf()` to convert existing R `data.frame`s
+  to their `polars` equivalents.
```
eitsupi, line 206:
```suggestion
  to their polars equivalents.
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+
+* The class names `PTime` and `rpolars_raw_list` (used to handle time and binary
+  variables) are removed. One should use the classes provided in packages
+  `hms` and `blob` instead.
```
eitsupi, line 210:
```suggestion
  hms and blob instead.
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  #> 1 12:00  <raw 1 B>
+  #> 2    NA         NA
+  #> 3 14:00  <raw 1 B>
+  ```
+
+
+### Other changes
+
+* R objects that convert to a Series of length 1 are now treated like scalar
+  values when converting to `polars` Expr:
```
eitsupi, line 278:
```suggestion
  values when converting to polars expressions:
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  #> ╞═════╪═════╡
+  #> │ 1   ┆ 1.0 │
+  #> │ 2   ┆ 1.0 │
+  #> └─────┴─────┘
+  ```
+
+* `map_batches()` still exists but its usage is discouraged. This function is
+  not guaranteed to interact correctly with the streaming engine. To apply
+  functions from external packages or custom functions that cannot be translated
+  to polars syntax, we now recommend converting the data to a `data.frame` and
```
eitsupi, line 310:
```suggestion
  to polars syntax, we now recommend converting the data to an R DataFrame and
```

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  using the parallel framework of `purrr` (>= 1.1.0) for efficiency. The vignette
```
eitsupi, line 311:
```suggestion
  using the parallel framework of purrr (>= 1.1.0) for efficiency. The vignette
```

I don't think it is a good idea to assume parallelism here.
Simply `lapply()` should be sufficient in many cases.

How about mentioning something like "parallelization is possible when using purrr"?

eitsupi (changes requested):

NEWS.md
etiennebacher: Now:

>  we now recommend converting the data to a `data.frame` and  using purrr (note that as of 1.1.0, purrr enables parallel computation).

NEWS.md
etiennebacher: Disagree on this, I don't remember seeing any other packages where this syntax is used.

NEWS.md
```diff
@@ -1,358 +1,315 @@
...
+  #> │ bar ┆ foo │
+  #> │ --- ┆ --- │
+  #> │ i32 ┆ f64 │
+  #> ╞═════╪═════╡
+  #> │ 1   ┆ 1.0 │
+  #> │ 2   ┆ 1.0 │
+  #> └─────┴─────┘
+  ```
+
+* `map_batches()` still exists but its usage is discouraged. This function is
```
eitsupi, line 307:
```suggestion
* `<expr>$map_batches()` still exists but its usage is discouraged. This function is
```

eitsupi (changes requested):

NEWS.md
eitsupi: Ok.

eitsupi (approved):
Thanks!

PR #290, opened 2025-04-11
feat: Add aggregate generic function S3 methods for Series

etiennebacher:


R/neopolars-package.R
```diff
@@ -4,5 +4,6 @@
 ## usethis namespace: start
 #' @import rlang
 #' @importFrom utils findMatches head
+#' @importFrom stats median
```
eitsupi, line 7: Is this needed?
How about `#' @exportS3Method stats::median`?

etiennebacher: Didn't know this one, fixed

R/series-s3-base.R
```diff
@@ -67,3 +67,18 @@ as.vector.polars_series <- function(x, mode = "any") {
...
+max.polars_series <- function(x, ...) x$max()
+
+#' @export
+min.polars_series <- function(x, ...) x$min()
+
+#' @export
+mean.polars_series <- function(x, ...) x$mean()
+
+#' @exportS3Method stats::median
+median.polars_series <- function(x, ...) x$median()
```
eitsupi, line 81: Maybe this function should be moved to `series-s3-stats.R`?

etiennebacher: It's part of base R so I think it's fine here, just like we don't have `series-s3-utils.R` for the dollar method.

eitsupi (approved):

PR #295, opened 2025-04-12
feat: Add `format()` S3 method for DataType

etiennebacher:


R/datatypes-s3-base.R
```diff
@@ -38,7 +38,12 @@
...
 #' @export
 print.polars_dtype <- function(x, ...) {
-  x$`_dt`$as_str(abbreviated = FALSE) |>
+  format(x, abbreviated = FALSE) |>
     writeLines()
   invisible(x)
 }
+
+#' @export
+format.polars_dtype <- function(x, abbreviated = FALSE, ...) {
```
eitsupi, line 47: - Doesn't the additional argument need to be after `...`?
- Can you add documentation to help us understand how to use it?

eitsupi (changes requested):
Thanks, could you also update here?
https://github.com/eitsupi/neo-r-polars/blob/c0e554ea2ce4c017aaa131fd98da0824b9e5389a/R/as_polars_df.R#L92

R/datatypes-s3-base.R
etiennebacher: Done, but since the other methods do not produce an Rd file, I don't think this one should too.

I've also added a check for empty dots.

etiennebacher:
> Thanks, could you also update here?

Done in [`6e133df` (#295)](https://github.com/eitsupi/neo-r-polars/pull/295/commits/6e133df266951e71ec4f3d509240fd7b3342b319) but I'm open to change the wording of the error message.

eitsupi (approved):
Thanks, I'll merge it anyway.

PR #254, opened 2025-03-31
build: Setup documentation website

etiennebacher:
- [x] syntax highlighting
- [x] reorder "Expressions" section in "Reference"
- [x] complete "altdoc/reference_home.Rmd"
- [x] create "IO" section for read/write/scan/sink
- [x] taskfile
- [x] ci
- [x] update "About" page
- [x] add a page like "reference_home.md" for Series to explain the dispatch Expr -> Series
- [x] add missing classes: GroupBy, ~RollingGroupBy~ etc. (requires #259). Actually functions of RollingGroupBy and DynamicGroupBy don't have a dedicated page, `agg()` is only mentioned on the `rolling()` and `group_by_dynamic()` pages.

PR #296, opened 2025-04-12
docs(readme): Rewrite to prepare for release

etiennebacher:
Just a copy-paste of the README currently used on the r-polars website + adds a small "Development" section at the end with the info from the former "About" page.


etiennebacher:
Can this be merged?

eitsupi:
I'm sorry, but I want to rewrite it, so please wait a while.

README.Rmd
```diff
@@ -15,75 +17,67 @@ knitr::opts_chunk$set(
...
+[![Docs dev version](https://img.shields.io/badge/docs-dev-blue.svg)](https://pola-rs.github.io/r-polars)
+<!-- badges: end -->
 
-I thought that now, around the release of Python Polars 1.0.0, is a good time for a complete rewrite, so I decided to try it.
+## Overview
 
-There are several reasons to rewrite r-polars on both the Rust and R sides.
+Polars is a DataFrame interface on top of an OLAP Query Engine implemented in Rust using
+[Apache Arrow Columnar Format](https://arrow.apache.org/docs/format/Columnar.html) as the memory
+model.
```
etiennebacher, line 35: I know this is from the README of Python Polars but to me it's just confusing.

I think it's very likely that an average R doesn't know what an OLAP query engine or the Arrow columnar format is (I don't even know what OLAP means). 

I don't think we should have this to start the README. The current r-polars README is more user-friendly IMO (although we don't have to list all kinds of operations Polars can perform).

README.Rmd
```diff
@@ -15,75 +17,67 @@ knitr::opts_chunk$set(
...
 
-### Rust side
+- Lazy | eager execution
+- Multi-threaded
+- SIMD
+- Query optimization
+- Powerful expression API
+- Hybrid Streaming (larger-than-RAM datasets)
+- Rust | Python | NodeJS | R | ...
```
etiennebacher, line 43: Similarly, I expect most average R users to not be knowledgeable about multithreading, SIMD, query optimization, etc.

README.Rmd
```diff
@@ -15,75 +17,67 @@ knitr::opts_chunk$set(
...
 
-### Current Status
+## Usage
 
-The directory structure on the Rust side is a complete copy of py-polars.
-The structure of R classes is also the same as py-polars.
+To avoid conflicts with other packages and base R function names, `{polars}`'s
+top level functions are hosted in the `pl` environment, and accessible via the
+`pl$` prefix.
+This means that `polars` queries written in Python and in R are very similar.
```
etiennebacher, line 66: Nitpick but we should decide if we use `polars` or `{polars}` in the README (and in the docs in general).

(I personally prefer `polars` just to avoid extra typing)

README.Rmd
```diff
@@ -15,75 +17,67 @@ knitr::opts_chunk$set(
...
 
-The basic classes such as `DataFrame`, `Series`, `Expr`, and `LazyFrame` have been implemented,
-and some functions work correctly.
+For example, rewriting the Python example from <https://github.com/pola-rs/polars> in R:
```
etiennebacher, line 68:
```suggestion
For example, rewriting the Python example from <https://github.com/pola-rs/polars#python> in R:
```

So that this goes straight to the example instead of the main page of the repo.

README.Rmd
```diff
@@ -15,75 +17,67 @@ knitr::opts_chunk$set(
...
+library(neopolars) # TODO: rename to polars
+
 df <- pl$DataFrame(
   A = 1:5,
   fruits = c("banana", "banana", "apple", "apple", "banana"),
   B = 5:1,
   cars = c("beetle", "audi", "beetle", "beetle", "beetle"),
 )
 
+# embarrassingly parallel execution & very expressive query language
```
etiennebacher, line 80:
```suggestion
# very expressive query language
```

It is embarrassingly parallel, but it's just invisible to the user so it feels weird to mention it here.

etiennebacher (commented):
Some comments below:

1. I think the example isn't very "attractive" to someone who discovers Polars for the first time. I'm not convinced mimicking the README of Python Polars is the way to go. The README is used as homepage for us, but they don't use their README as their homepage (https://docs.pola.rs/). The scan + filter + agg + collect on their homepage is an easier-to-read example IMO. Then if we want to show the similarity between the R and Python implementations, we can have a small code block showing some code in Python followed by a small code block showing the same in R.
2. I'd like to keep the "Development" section that was added in one of the first commits of this branch (cf https://github.com/eitsupi/neo-r-polars/pull/254#discussion_r2034158312). Maybe it's what you intend for "Governance"?

README.Rmd
eitsupi: I understand what you are saying, but I don't know if the terminology difficulty justifies leaving this place, given that PyPI[^pypi], crates.io[^cratesio], or npm[^npm] all show something like this in their READMEs.

[^pypi]: <https://pypi.org/project/polars/>
[^cratesio]: <https://crates.io/crates/polars>
[^npm]: <https://www.npmjs.com/package/nodejs-polars>

README.Rmd
eitsupi: Only use `` `polars` `` now.

README.Rmd
etiennebacher: I don't think an average user goes to either of those sites. They go straight to the official documentation website which is much more user friendly (https://docs.pola.rs/).

I really think we should drop this paragraph and keep the one on current r-polars: it has no value added for the average user, and the experienced ones can easily find this info elsewhere.

eitsupi: I see. How about I copied it from the index page of the user guide?

etiennebacher (commented):
> How about I copied it from the index page of the user guide?

Thanks, I think it's better. I still have a few suggestions though. 

Except for the functions reference, I think it's ok to deviate from Polars docs and not have a complete 1-to-1 mapping.

README.Rmd
```diff
@@ -15,153 +17,145 @@ knitr::opts_chunk$set(
...
+Polars R bindings.
 
-There are several reasons to rewrite r-polars on both the Rust and R sides.
+Polars is a blazingly fast DataFrame library for manipulating structured data. The core is written
+in Rust, and available for Python, R and NodeJS.
 
-### Rust side
+- **Fast**: Written from scratch in Rust, designed close to the machine and without external
+  dependencies.
+- **I/O**: First class support for all common data storage layers: local & cloud storage.
```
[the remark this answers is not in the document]
eitsupi: I think this works after #354, but Polars cannot read from public blob storage without credentials, so without an account I cannot test.
(Checking the upstream object-store, it looks like there is functionality for anonymous access, if the upstream Polars is modified, anonymous access should be allowed, but I didn't have time to create a bug report.)

eitsupi:
@etiennebacher Thanks for your review.
I have updated it to take your suggestion into consideration, how do you like it?

README.Rmd
```diff
@@ -15,153 +17,126 @@ knitr::opts_chunk$set(
...
+[![CRAN status](https://www.r-pkg.org/badges/version/polars)](https://CRAN.R-project.org/package=polars)
+[![Docs dev version](https://img.shields.io/badge/docs-dev-blue.svg)](https://pola-rs.github.io/r-polars)
+<!-- badges: end -->
 
-I thought that now, around the release of Python Polars 1.0.0, is a good time for a complete rewrite, so I decided to try it.
+Polars is a blazingly fast DataFrame library for manipulating structured data.
+The core is written in Rust.
 
-There are several reasons to rewrite r-polars on both the Rust and R sides.
+The goal of Polars is to provide a lightning fast DataFrame library that:
```
etiennebacher, line 34: The "ligthning fast" and "blazingly fast" are duplicated here. How about:

```
*badges*

Polars is a blazingly fast DataFrame library, written in Rust. Its goal
is to deliver fast, efficient data processing that:
```

README.Rmd
```diff
@@ -15,153 +17,126 @@ knitr::opts_chunk$set(
...
 
-### Rust side
+- Utilizes all available cores on your machine.
+- Optimizes queries to reduce unneeded work/memory allocations.
+- Handles datasets much larger than your available RAM.
+- A consistent and predictable API.
```
etiennebacher, line 39:
```suggestion
- Follows a consistent and predictable API.
```

README.Rmd
```diff
@@ -15,153 +17,126 @@ knitr::opts_chunk$set(
...
-   it is not possible to place multiple impl blocks.
-   ([extendr/extendr#538](https://github.com/extendr/extendr/issues/538))
-2. Error handling.
-   There is a lot of custom code to use the Result type with extendr, which is quite different from other packages based on extendr.
-   ([extendr/extendr#650](https://github.com/extendr/extendr/issues/650))
-3. Simplify type conversion.
-   The code is difficult to follow because it uses a macro called `robj_to` for type conversion (at least in rust-analyzer).
+This `polars` R package provides the R bindings for Polars.
+It can be used to convert R DataFrames to Polars DataFrames and vice versa,
+as well as to integrate with other common R packages.
```
etiennebacher, line 44: I feel it's a bit reductive to mention only conversion between R and Polars. How about this?

```suggestion
It can be used to process Polars DataFrames and LazyFrames, convert objects 
between Polars and R, and can be integrated with other common R packages.
```

etiennebacher (commented):
I like it, I have a few other comments I missed in the previous review (sorry).

README.Rmd
eitsupi: Again, I'm not sure if the term LazyFrame is appropriate here; I believe LazyFrame is commonly referred to as a Query Plan or something similar.

eitsupi: How about 54ad963d1502264065b3d13a8cabddf5eddd2623?

etiennebacher: It appears many times in the user guide, even when introducing the lazy api: https://docs.pola.rs/user-guide/lazy/

There's even a function called LazyFrame so I don't see a problem with that.

etiennebacher: > How about [54ad963](https://github.com/eitsupi/neo-r-polars/commit/54ad963d1502264065b3d13a8cabddf5eddd2623)?

Fine for  me, I just don't want to replace all "LazyFrame" in docs by query plans.

PR #297, opened 2025-04-12
build: Add CI workflow to check that all man pages are in `mkdocs.yml`

etiennebacher:
Since the generation of `mkdocs.yml` is no longer automatic, it is not guaranteed anymore that all functions are listed there (it is easy to forget this step when adding a new function).

This PR adds a workflow to detect that all man pages are listed in `mkdocs.yml`.

 ~CI failures are expected, needs to be rebased on #298.~

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -1,0 +1,57 @@
+name: Check missing docs
+
+on:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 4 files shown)


eitsupi (changes requested):
Can we change the file name to something like "test-docs.yml" and the workflow name to something like "Test documentation" to match other workflows?

etiennebacher:
Sure, done

eitsupi (approved):

PR #302, opened 2025-04-13
fix: Some functions for Series must return an R value

### LINKED ISSUE #289: Some Series functions must return an R value
E.g. `<Series>$max()`:

```r
polars::pl$Series("", 1:2)$max()
# 2

neopolars::pl$Series("", 1:2)$max()
# shape: (1,)
# Series: '' [i32]
# [
#	 2
# ]
```

```python
import polars as pl
pl.Series([1, 2]).max()
# 2
```

Functions where this is the case: mean, median, min, max, sum (maybe more)

etiennebacher:
Fixes #289


eitsupi (commented):
This implementation is interesting, but looking at the Python Polars implementation, I think it would be better to do the same since it uses dedicated functions on the Rust side.

https://github.com/pola-rs/polars/blob/cd3e4ee7995f95bfb186d6f9b323631db246df60/crates/polars-python/src/series/aggregation.rs#L45-L47

src/rust/src/series/aggregation.rs
```diff
@@ -0,0 +1,111 @@
...
+use polars::prelude::{QuantileMethod, Scalar};
+use polars_core::prelude::AnyValue;
+use savvy::{NumericScalar, Result, Sexp, savvy};
+
+fn scalar_to_r(x: Scalar) -> Result<Sexp> {
+    match x.as_any_value() {
+        AnyValue::Int8(x) => return (x as i32).try_into(),
+        AnyValue::Int16(x) => return (x as i32).try_into(),
+        AnyValue::Int32(x) => return x.try_into(),
+        AnyValue::Int64(x) => return (x as f64).try_into(),
```
etiennebacher, line 12: `product()` returns i64 but I don't think we want to add the arg `int64_conversion` in this function only (but we could).

etiennebacher:
I've added functions in Rust, there were also several more that I missed in the first iteration.

eitsupi (commented):
Thank you. This seems to be more work than I thought.

Thinking about it again, I'm wondering if it's not worth returning values as a class of R instead of a class of Polars.

- Base R classes may be missing information (ex. Int64 -> double).
- Since there is no distinction between scalars and vectors here, we can immediately obtain a vector of length 1 by performing the general operation of converting Series to R later.

etiennebacher:
> I'm wondering if it's not worth returning values as a class of R instead of a class of Polars.

I don't understand what you mean, can you clarify? 

> Base R classes may be missing information (ex. Int64 -> double).

The first way I implemented this (by calling `to_r_vector()` if one of those functions is used) would allow the user to control the way int64 are returned via the `int64_conversion` global option.

eitsupi:
> > I'm wondering if it's not worth returning values as a class of R instead of a class of Polars.
> 
> I don't understand what you mean, can you clarify?

This means that returning a Series of length 1 may be more valuable than returning a vector of length 1.
Why is it necessary for a vector to be the return value here?

etiennebacher:
>  Why is it necessary for a vector to be the return value here?

Well for consistency with python-polars and also because then we want to call those in various S3 methods such as `sum()`, `max()`, etc. (cf #290) and it would be better for consistency if those returned a scalar. 

I think what I put above is still valid:

> The first way I implemented this (by calling to_r_vector() if one of those functions is used) would allow the user to control the way int64 are returned via the int64_conversion global option.

eitsupi:
> > Why is it necessary for a vector to be the return value here?
> 
> Well for consistency with python-polars and also because then we want to call those in various S3 methods such as `sum()`, `max()`, etc. (cf #290) and it would be better for consistency if those returned a scalar.

In Python, Series corresponds to list, which is justified in that what an aggregate function returns should not be a list. This is same to `numpy.array`.

In R, Series corresponds to a vector, and the aggregate function must return a vector. In other words, it is not a problem to aggregate a Series to become a Series.

In addition, I don't think that generic functions such as `mean` always need to return atomic vectors.
In fact, aggregating `arrow::Array` returns `arrow::Scalar`.

``` r
arrow::as_arrow_array(1:10) |> mean()
#> Scalar
#> 5.5
```

<sup>Created on 2025-04-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher:
>  In fact, aggregating arrow::Array returns arrow::Scalar.

Fair enough, but I don't think this is necessarily a good choice. In terms of usability, returning a Series means that the output cannot be used directly in any other R function as-is, it would need to go through `as.vector()`. This is in contrast to the python-polars that can also be used in any other python function and not only those from polars.

eitsupi:
> In terms of usability, returning a Series means that the output cannot be used directly in any other R function as-is

What are the specific uses?
I am concerned that when exporting R values from a Series, accuracy may be lost, potentially compromising its value for downstream use.

In addition, if you wish to convert to R values on the Rust side, please be sure to go through `PlRSeries::to_r_vector` instead of implementing a new conversion on your own.

etiennebacher:
>  I am concerned that when exporting R values from a Series, accuracy may be lost, potentially compromising its value for downstream use.

Ok I see that this is an issue. Let's close this.

eitsupi:
I recall where the aggregate functions in Series need to return a "scalar", since Polars does not broadcast a Series of length 1, e.g. `with_columns`, the following error says _use `.first()` to get a scalar_ would occur.

``` r
library(neopolars)

pl$DataFrame(a = 1:3)$with_columns(
  b = pl$Series("", 1)
)
#> Error:
#> ! Evaluation failed in `$with_columns()`.
#> Caused by error:
#> ! Evaluation failed in `$collect()`.
#> Caused by error:
#> ! Invalid operation: Series b, length 1 doesn't match the DataFrame height of 3
#>
#> If you want expression: Series to be broadcasted, ensure it is a scalar (for instance by adding '.first()').

pl$DataFrame(a = 1:3)$with_columns(
  b = pl$Series("", 1)$first()
)
#> Error:
#> ! Evaluation failed in `$with_columns()`.
#> Caused by error:
#> ! Evaluation failed in `$collect()`.
#> Caused by error:
#> ! Invalid operation: Series b, length 1 doesn't match the DataFrame height of 3
#>
#> If you want expression: Series to be broadcasted, ensure it is a scalar (for instance by adding '.first()').
```

<sup>Created on 2025-04-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In neopolars, the reason why passing an ordinary R-vector instead of a Series here does not cause an error is that special handling is performed for length 1 in the following section.

https://github.com/eitsupi/neo-r-polars/blob/6678facf79a5677fe55ca83b4fcfd86891a18a6e/R/as_polars_expr.R#L122-L125

So far, I have not done the above special processing for Series of length 1 as in Python, but perhaps the above processing is necessary for consistency with R vector.

In Python, the aggregate function returns a Python scalar, so `first` works as follows.

```python
>>> import polars as pl
>>> pl.DataFrame({"a": [1, 2, 3,]}).with_columns(
...   b = pl.Series("", [1])
... )
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/rstudio/.local/lib/python3.12/site-packages/polars/dataframe/frame.py", line 9806, in with_columns
    return self.lazy().with_columns(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rstudio/.local/lib/python3.12/site-packages/polars/_utils/deprecation.py", line 93, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rstudio/.local/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 2206, in collect
    return wrap_df(ldf.collect(engine, callback))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.InvalidOperationError: Series b, length 1 doesn't match the DataFrame height of 3

If you want expression: Series to be broadcasted, ensure it is a scalar (for instance by adding '.first()').

>>> pl.DataFrame({"a": [1, 2, 3,]}).with_columns(
...   b = pl.Series("", [1]).first()
... )
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i32 │
╞═════╪═════╡
│ 1   ┆ 1   │
│ 2   ┆ 1   │
│ 3   ┆ 1   │
└─────┴─────┘
```

PR #304, opened 2025-04-14
feat: Support serialization and deserialization for LazyFrame

eitsupi:


R/lazyframe-frame.R
etiennebacher, line 1: FYI there is a TODO to remove at the bottom of this file

PR #305, opened 2025-04-14
ci: Miniaturization of the `R CMD check` test matrix

eitsupi:
I think the oldest and newest versions supported here are sufficient.

### DIFF
--- a/.github/workflows/test-r.yml
+++ b/.github/workflows/test-r.yml
@@ -54,7 +54,6 @@ jobs:
         os:
           - ubuntu-latest
         r:
-          - release
           - devel
           - "4.3"
     steps:


eitsupi:
As a matter of fact, R release is tested here with the debug profile on macos and the release profile on windows.
https://github.com/eitsupi/neo-r-polars/blob/f66d68fc3768a3ce94dfcd4acd8827c195e2e63e/.github/workflows/test-r.yml#L119-L124

R CMD check is just tests to pick up bugs that are not found by mere testing, so R devel, which is considered more stringent, is sufficient Since R CMD check takes a long time to run, I would like to reduce the number of jobs to avoid the limitation of parallel execution.

etiennebacher:
Ok, I agree we probably don't want to test all combinations of R version * platform * {profile, debug}, except maybe on PRs made right before a new release?

eitsupi:
Yes, I would do all builds with the "dest-release" profile with optimizations (same as Python Polars) before releasing the Rust library.

PR #310, opened 2025-04-17
feat: Add `polars_options()`

### LINKED ISSUE #282: About `polars_options()`
Looking at the current `polars` options:

``` r
polars::polars_options()
#> Options:
#> ========                         
#> debug_polars        FALSE
#> df_knitr_print       auto
#> do_not_repeat_call  FALSE
#> int64_conversion   double
#> limit_max_threads    TRUE
#> maintain_order      FALSE
#> no_messages         FALSE
#> rpool_active            0
#> rpool_cap               4
#> strictly_immutable   TRUE
#> 
#> See `?polars_options` for the definition of all options.
```

A lot of them look outdated for the rewritten package. Actually, it looks like the only useful ones are `int64_conversion` and maybe `maintain_order` (but I'm not even convinced of that).

@eitsupi what do you think? Do you see more that are actually useful?

etiennebacher:
Fixes #282


R/polars_options.R
```diff
@@ -0,0 +1,84 @@
...
+#'
+#' * `int64_conversion` (`"double"`): How should Int64 values be handled when
+#'   converting a polars object to R?
+#'    * `"double"` converts the integer values to double.
+#'    * `"integer"` converts to the R's [integer] type. If the value is out of
+#'      the range of R's integer type, export as [NA_integer_].
+#'    * `"integer64"` uses `bit64::as.integer64()` to do the conversion
+#'      (requires the package `bit64` to be attached).
+#'    * `"character"` converts Int64 values to character.
+#' * `df_knitr_print` (TODO: possible values??)
```
etiennebacher, line 24: I don't see any usage of `df_knitr_print` in the package so far, so leaving this to be completed

eitsupi (changes requested):
Sorry, but I really don't have time to write down my thoughts, which are not coherent.

- The name `polars.int64_conversion` must be changed something like `polars.to_r_vector.int64` (Since we need to have options for all other arguments)
- Note that any R object may be assigned here.
- Shouldn't the default value be `NULL`? In that case there should be no need for something like the comparison with `"double"` as you have currently written.

eitsupi:
So what I am saying is that comparisons using the `==` operator should not be done inside `if ()` because it is not guaranteed to be a string.

I think we need something like this:

``` r
library(rlang)

f1 <- \(x = c("foo", "bar")) {
  if (missing(x)) {
    from_opt <- getOption("is_missing", x[1])
    if (!identical(from_opt, x[1])) {
      inform(
        sprintf(
          '`x` is overridden by the option "is_missing" with %s',
          rlang:::obj_type_friendly(from_opt)
        ),
      )
      x <- from_opt
    }
  }
  arg_match0(x, c("foo", "bar"))
}

f1()
#> [1] "foo"
f1("bar")
#> [1] "bar"

withr::with_options(
  list(is_missing = "foo"),
  f1()
)
#> [1] "foo"
withr::with_options(
  list(is_missing = "bar"),
  f1()
)
#> `x` is overridden by the option "is_missing" with the string "bar"
#> [1] "bar"
withr::with_options(
  list(is_missing = 1L),
  f1()
)
#> `x` is overridden by the option "is_missing" with the number 1
#> Error in `f1()`:
#> ! `x` must be a string or character vector.
```

<sup>Created on 2025-04-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/polars_options.R
```diff
@@ -0,0 +1,105 @@
...
+polars_options <- function() {
+  out <- list(
+    df_knitr_print = getOption("polars.df_knitr_print") %||% "auto",
+    conversion_int64 = getOption("polars.conversion_int64") %||% "double",
+    conversion_uint8 = getOption("polars.conversion_uint8") %||% "integer",
+    conversion_date = getOption("polars.conversion_date") %||% "Date",
+    conversion_time = getOption("polars.conversion_time") %||% "hms",
+    conversion_decimal = getOption("polars.conversion_decimal") %||% "double",
+    conversion_ambiguous = getOption("polars.conversion_ambiguous") %||% "raise",
+    conversion_non_existent = getOption("polars.conversion_non_existent") %||% "raise"
```
eitsupi, line 42: I don't think the word "conversion" should be used because it is not clear what is meant by conversion from something to something else.

eitsupi (changes requested):

R/polars_options.R
etiennebacher: I put `to_r_vector`, e.g. `polars.to_r_vector_int64` but I'm also open to `polars.to_r_int64` to have shorter names.

etiennebacher:
> So what I am saying is that comparisons using the == operator should not be done inside if () because it is not guaranteed to be a string.

Ah I see, I've updated that.

etiennebacher:
One test fails:

```r
# can convert to bit64, but *only* if bit64 is attached
  try(detach("package:bit64"), silent = TRUE)
  withr::with_options(
    list(polars.to_r_vector_int64 = "integer64"),
    expect_snapshot(as.list(df, as_series = FALSE), error = TRUE)
  )
```

Currently, this doesn't error. In this PR, I check that necessary packages are *loaded* when printing `polars_options()`. However, in `to_r_vector()` (called by `as.list()`), we only check whether the required packages are *installed*, which is the case here and the reason this check fails. Should we check if required packages are loaded or installed?

R/series-s3-base.R
```diff
@@ -59,7 +59,8 @@ METHODS_EXCLUDE <- c(
 # TODO: support the mode argument
 #' @export
 as.vector.polars_series <- function(x, mode = "any") {
-  x$to_r_vector(ensure_vector = TRUE) |>
+  int64 <- use_option_if_missing(int64, TRUE, "double")
+  x$to_r_vector(ensure_vector = TRUE, int64 = int64) |>
```
eitsupi, line 63: Is this needed?

R/utils-various.R
```diff
@@ -19,3 +19,49 @@ extend_bool <- function(value, n_match, value_name, match_name) {
...
+        "` must be attached to use `",
+        option_name,
+        " = \"",
+        option_value,
+        "\"`."
+      ),
+      call = caller_env()
+    )
+  }
+}
```
eitsupi, line 67: Move these to `polars_options.R`, as I believe this is where we would put a copy of the Python Polars functions.

eitsupi (changes requested):

eitsupi:
I do remember that something `hms` etc. was not ideal because it would print as `difftime` if not loaded even if already installed.
But could that be left for now as room for future improvement?

R/series-s3-base.R
etiennebacher: Yes, to ensure that the global option is correctly used in `as.vector()`, e.g. in this test:

```r
withr::with_options(
    list(polars.to_r_vector_int64 = "character"),
    {
      expect_snapshot(
        expect_identical(
          as.vector(pl$Series("a", c(1:3, NA))$cast(pl$Int64)),
          c("1", "2", "3", NA)
        )
      )
})
```

R/utils-various.R
etiennebacher: Done

etiennebacher:
> But could that be left for now? Concentrate on options here, loading, etc. can be improved later.

I can't reproduce the weird failures on macOS but the other tests now pass, I've added a TODO for the previously failing one. Let me know if you'd rather keep only the changes in `polars_options.R` in this PR and move all the changes related to handling global options in `as.` functions to another PR.

eitsupi:
> I do remember that something `hms` etc. was not ideal because it would print as `difftime` if not loaded even if already installed.

Since `rlang::is_installed()` loads the package, it is not necessary for the external dependency package to have been loaded at the time `$to_r_vector()` is called.
`vctrs`, `blob`, and `hms` will be reliably loaded after #323.

They do not need to be attached. Those checks and tests should be removed.

etiennebacher:
I have updated tests and removed some checks but I'm not sure this is what you had in mind, can you check?

Also, what do you think of renaming options, e.g. `polars.to_r_vector_int64` to `polars_to_r_int64`?

eitsupi:
It seems that `rlang::missing_arg()` would simplify implementation.

``` r
library(rlang)

f1 <- \(x = "foo") {
  if (missing(x)) {
    x <- getOption("is_missing", "foo")
  }
  x
}

f2 <- \(x = "foo") {
  if (missing(x)) {
    x <- missing_arg()
  }

  f1(x = x)
}

f1()
#> [1] "foo"
f2()
#> [1] "foo"

withr::local_options(list(is_missing = "overridden"))

f1()
#> [1] "overridden"
f1("bar")
#> [1] "bar"
f2()
#> [1] "overridden"
f2("bar")
#> [1] "bar"
```

<sup>Created on 2025-05-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

> what do you think of renaming options, e.g. `polars.to_r_vector_int64` to `polars.to_r_int64`?

I do not like these option names. To me, it looks like it specifies _something that will be converted to R int64_.
It is actually an option to specify _something to be converted from int64_.

etiennebacher:
> It seems that rlang::missing_arg() would simplify implementation.

Sorry, I don't see anywhere in the code where this could be applied, can you maybe indicate where you'd use it?

>  It is actually an option to specify something to be converted from int64.

I've renamed those to be `polars.from_int64`, `polars.from_date`, etc. Please let me know if that's fine for you and propose names if it's not.

eitsupi:
> Sorry, I don't see anywhere in the code where this could be applied, can you maybe indicate where you'd use it?

I think it would be simpler to refer to options only within the `series__to_r_vector()` and simply pass `missing_arg()` in downstream functions, rather than referring to options inside all functions.

> I've renamed those to be `polars.from_int64`, `polars.from_date`, etc. Please let me know if that's fine for you and propose names if it's not.

See https://github.com/eitsupi/neo-r-polars/pull/310#pullrequestreview-2775970813
`polars.from_int64` feels to me like a mapping of the conversion from int64 in R.
I prefer something like `polars.to_r_vector.int64` because it is clearly the `int64` argument of the `to_r_vector` function.

R/polars_options.R
```diff
@@ -1,3 +1,4 @@
+# TODO: add engine_affinity
```
etiennebacher, line 1: What is engine affinity?

R/polars_options.R
```diff
@@ -104,17 +111,17 @@ print.polars_options <- function(x, ...) {
 #' @param is_missing Is `x` missing in the calling function?
 #' @param default The default of `x` in the calling function
 #' @noRd
-use_option_if_missing <- function(x, is_missing, default) {
+use_option_if_missing <- function(x, is_missing, default, prefix = "polars.") {
```
etiennebacher, line 114: I don't understand this change, is there any option we don't want to prefix with `polars.`? This adds noise in `series__to_r_vector`. We could just define `prefix <- "polars.to_r_vector."` at the top of this function.

R/polars_options.R
eitsupi: pola-rs/polars#20717

In Python, it is specified by an environment variable, but in R, I think it might be useful to specify it by options. This is because Python and R have different engines that can be used, so if we specify an environment variable that is valid in Python, it may not work in R.

etiennebacher: Thanks

eitsupi (approved):

R/dataframe-s3-base.R
```diff
@@ -55,6 +55,34 @@ as.list.polars_data_frame <- function(
...
+  if (missing(as_clock_class)) {
+    as_clock_class <- missing_arg()
+  }
+  if (missing(ambiguous)) {
+    ambiguous <- missing_arg()
+  }
+  if (missing(non_existent)) {
+    non_existent <- missing_arg()
+  }
+
```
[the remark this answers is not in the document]
eitsupi: IIUC, it does not work.

tests/testthat/test-polars_options.R
```diff
@@ -0,0 +1,119 @@
...
+      "polars.to_r_vector.decimal", "character",
+      "polars.to_r_vector.as_clock_class", FALSE,
+      "polars.to_r_vector.as_clock_class", TRUE,
+      "polars.to_r_vector.ambiguous", "raise",
+      "polars.to_r_vector.ambiguous", "earliest",
+      "polars.to_r_vector.ambiguous", "latest",
+      "polars.to_r_vector.ambiguous", "null",
+      "polars.to_r_vector.non_existent", "raise",
+      "polars.to_r_vector.non_existent", "null",
+    )[-1, ] # remove the first dummy row
```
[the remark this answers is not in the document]
eitsupi: Please check the comment. It is needed for to make the column as list.

Issue #301, opened 2025-04-13
Document about `mirai` integration


eitsupi:
The functionality that exists in the current `polars` to launch an R session from the Rust side and evaluate it in the background is intentionally not implemented.
In my opinion, it is more versatile and simpler to use the `mirai` package and we need to encourage people to use it.

For example, an example using map_batches:

``` r
library(polars)

pl$LazyFrame(a = 1, b = 2, c = 3, d = 4)$select(
  pl$all()$map_batches(\(s) {
    Sys.sleep(.1)
    s * 2
  }, in_background = TRUE)
)$collect()
#> shape: (1, 4)
#> ┌─────┬─────┬─────┬─────┐
#> │ a   ┆ b   ┆ c   ┆ d   │
#> │ --- ┆ --- ┆ --- ┆ --- │
#> │ f64 ┆ f64 ┆ f64 ┆ f64 │
#> ╞═════╪═════╪═════╪═════╡
#> │ 2.0 ┆ 4.0 ┆ 6.0 ┆ 8.0 │
#> └─────┴─────┴─────┴─────┘
```

<sup>Created on 2025-04-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This can be done the same way with `mirai` after #300, as follows. Note that they are run in parallel.
(The Int128 type is used to indicate that the Series have been correctly restored.)

``` r
library(neopolars)

cfg <- mirai::serial_config(
  class = "polars_series",
  sfun = \(x) x$serialize(),
  ufunc = pl$deserialize_series
)

mirai::daemons(4, serial = cfg)
#> [1] 4

pl$LazyFrame(a = 1, b = 2, c = 3, d = 4)$cast(pl$Int128) |>
  as.list() |>
  mirai::mirai_map(\(s) {
    Sys.sleep(.1)
    s * 2L
  }) |>
  _[] |>
  as_polars_df()
#> shape: (1, 4)
#> ┌──────┬──────┬──────┬──────┐
#> │ a    ┆ b    ┆ c    ┆ d    │
#> │ ---  ┆ ---  ┆ ---  ┆ ---  │
#> │ i128 ┆ i128 ┆ i128 ┆ i128 │
#> ╞══════╪══════╪══════╪══════╡
#> │ 2    ┆ 4    ┆ 6    ┆ 8    │
#> └──────┴──────┴──────┴──────┘
```

[...]

Issue #151, opened 2025-02-25
Functions that are hard to implement


etiennebacher:
I don't know how to implement the following functions:

- [ ] ~map_batches ([py-polars](https://github.com/pola-rs/polars/blob/28cc37ad607eff376ef462cc9ce98f26c216e84f/crates/polars-python/src/map/lazy.rs#L163-L176))~ Convert to R and use `purrr` or `mirai` instead
- [ ] ~map_elements ([py-polars](https://github.com/pola-rs/polars/blob/28cc37ad607eff376ef462cc9ce98f26c216e84f/crates/polars-python/src/series/map.rs))~ Convert to R and use `purrr` or `mirai` instead
- [ ] arr_to_struct ([py-polars](https://github.com/pola-rs/polars/blob/28cc37ad607eff376ef462cc9ce98f26c216e84f/crates/polars-python/src/expr/array.rs#L112-L130))
- [ ] list_to_struct ([py-polars](https://github.com/pola-rs/polars/blob/28cc37ad607eff376ef462cc9ce98f26c216e84f/crates/polars-python/src/expr/list.rs#L211-L237))

All of those functions rely on calling a user-defined function from Rust. Is this possible with `savvy`? In `extendr` we could use the `.as_function()` method on a `Robj`: https://github.com/pola-rs/r-polars/blob/f58489cd69af3bb77fe7fcdcbcd5bf2383a2a77d/src/rust/src/series.rs#L365

etiennebacher:
Forget about this, I'll try to adapt the code in `map_single()` and reopen if I can't manage.

eitsupi:
These can be implemented, but seem to read the wrong memory location and crash, as in #5.

Also, `flume` currently used is in maintenance mode.
I do not recommend creating anything based on these, as they appear to need fundamental modifications.

eitsupi:
I think we need to recommend the use of something like regular R map functions instead of the `map_batches` etc.

Behavior in streaming mode may not be intuitive.
I need to recommend the use of perhaps something like `mirai::mirai_map`.

``` r
library(neopolars)

polars_info()
#> Polars R package version : 0.0.0.9000
#> Rust Polars crate version: 0.46.0
#>
#> Thread pool size: 16
#>
#> Features:
#> nightly FALSE

query <- pl$LazyFrame(
  a = 1:50,
  b = 1:50,
)$select(
  pl$all()$map_batches(\(x) {
    cat("run\n")
    x$to_r_vector() |>
[...]

PR #313, opened 2025-04-18
chore: Remove some outdated TODO comments

etiennebacher:


R/expr-expr.R
```diff
@@ -3992,20 +3992,6 @@ expr__gather_every <- function(n, offset = 0) {
...
-# #' @param fmt How to format the expression. Use `"{}"` where you want to print
-# #' the value of the expression.
-# #'
-# #' @inherit as_polars_expr return
-# #' @examples
-# #' df <- pl$DataFrame(foo = c(1, 1, 2))
-# #' df$select(pl$col("foo")$cum_sum()$inspect("value is: {}")$alias("bar"))
-# expr__format <- function(fmt) {
-#
-# }
```
etiennebacher, line 4007: Since `$map_batches()` will likely not be included (cf #301) there's no need to keep this.

R/dataframe-frame.R
```diff
@@ -432,8 +432,6 @@ dataframe__with_columns_seq <- function(...) {
     wrap()
 }
 
-# TODO-REWRITE: before release, add in news that param idx was renamed "index"
-# and mention that it errors if out of bounds
```
etiennebacher, line 436: So many breaking changes that it doesn't make sense to keep this TODO

R/expr-expr.R
eitsupi: Just to be clear, `<expr>$map_batches()` is included at this time. I merely question its usefulness.
And, it is still buggy (https://github.com/eitsupi/neo-r-polars/issues/151#issuecomment-2797139057).

eitsupi (approved):

R/expr-expr.R
etiennebacher: Right, I meant it like if we want to orient users to `purrr` / `mirai`, let's not use `map_batches()` too much internally.

PR #318, opened 2025-04-21
feat: Warn if `group_by()` receives an value named `maintain_order`

### LINKED ISSUE #283: Warn if user passes `maintain_order` in `group_by()`?
I think one of the biggest footguns with the new version is the added `.` as prefix of some argument names, and in particular the change from `maintain_order` to `.maintain_order`. Passing `maintain_order = TRUE` is valid in `group_by()` so this can go undetected.

How about we add a warning if the user passes `maintain_order = TRUE/FALSE` in the dynamic dots?

etiennebacher:
Closes #283

### DIFF
--- a/R/dataframe-group_by-general.R
+++ b/R/dataframe-group_by-general.R
@@ -2,6 +2,11 @@
 polars_groupby__methods <- new.env(parent = emptyenv())
 
 wrap_to_group_by <- function(x, by, maintain_order) {
+  if (any(names(by) == "maintain_order")) {
+    warn(
+      "In `$group_by()`, `...` contain an argument named `maintain_order`. Did you mean `.maintain_order` instead?"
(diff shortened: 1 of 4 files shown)


R/dataframe-group_by-general.R
```diff
@@ -2,6 +2,11 @@
 polars_groupby__methods <- new.env(parent = emptyenv())
 
 wrap_to_group_by <- function(x, by, maintain_order) {
+  if (any(names(by) == "maintain_order")) {
```
eitsupi, line 5: How about use `rlang::has_name()`?

etiennebacher: Thanks, done

etiennebacher (commented):
Thanks, LGTM

PR #321, opened 2025-04-23
docs: Add vignette on using custom functions and `purrr` / `mirai`

### LINKED ISSUE #301: Document about `mirai` integration

etiennebacher:
Fixes #301 

Just a quick draft to get the ball rolling. @eitsupi feel free to modify directly (if you have time of course, no pressure).

### DIFF
--- a/vignettes/custom-functions.Rmd
+++ b/vignettes/custom-functions.Rmd
@@ -1,0 +1,115 @@
+---
+title: "Using custom functions or other R packages"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Using custom functions or other R packages}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+```{r setup}
+library(neopolars)
+```
+
+`polars` contains a large set of functions to manipulate variables, whether they are numerical, strings, dates, or other types.
+Still, it is sometimes necessary to apply custom functions either written by you or available in other R packages.
+This vignette details several ways to do that in the most computationally efficient way.
+
+## Writing functions using `polars` expressions
+
+One of `polars`' biggest strength is the composability of expressions.
+It is very easy to chain multiple functions to create custom functions.
+For example, `polars` doesn't provide a function to standardize a variable out-of-the-box, but we could definitely create one using expressions:
+
+```{r}
+pl_standardize <- function(x) {
+  (x - x$mean()) / x$std()
+}
+```
+
+And we can then run it in `$with_columns()` for instance:
+```{r}
+dat <- as_polars_df(mtcars[, c("carb", "mpg", "drat")])
+dat$with_columns(
+  carb_stand = pl_standardize(pl$col("carb"))
+)
+```
+
+## Using `purrr`
+
+`polars` expressions allow one to write their own functions, but it is sometimes necessary to use functions that already exist in other R packages.
+When this is the case, one should keep in mind that applying those functions that are external to `polars` will limit `polars`' strengths.
+In particular, `polars` is fast in part because it is able to detect the various transformations applied to the data and run them in parallel or cache them so that only the strict minimum of computations is made.
+Using `polars` expressions will also allow the code to be run in streaming mode, allowing for larger-than-memory data.
+When one uses functions that `polars` doesn't natively know, those benefits are lost.
+This is the reason why it is better to use `polars` expressions in your own functions if you can.
+
+To apply an external function to a `polars` object, one must convert this object to R, apply the function, and then convert it back to `polars`.
+Note that when you convert your `polars` object to R, you may run out of memory if your object is a LazyFrame that is too big to fit in memory once collected.
+Another crucial point is that we want the R function to run as fast as possible on the data.
+One way to do that is to take advantage of `purrr` to run the function [in parallel](https://purrr.tidyverse.org/dev/reference/parallelization.html).
+
+**Note:** at the time of writing, this requires the development version of `purrr` (1.0.4.9000).
+
+```{r}
+# Set the number of background processes to use
+mirai::daemons(3)
+
+dat |>
+  as.list(as_series = FALSE) |>
+  purrr::map(
+    \(s) {
+      stats::rnorm(s)
+    },
+    .parallel = TRUE
+  ) |>
+  _[] |>
+  as_polars_df()
+```
+
+Note that this applies the function on the entire DataFrame (or LazyFrame).
+To apply this to a subset of columns only, you can use `map_at()`:
+
+```{r}
+dat |>
+  as.list(as_series = FALSE) |>
+  purrr::map_at(
+    c("carb", "mpg"),
+    \(s) {
+      stats::rnorm(s)
+    },
+    .parallel = TRUE
+  ) |>
+  _[] |>
+  as_polars_df()
+```
+
+
+## Using `mirai`
+
+`mirai` is the package that is used by `purrr` in the background to run functions in parallel.
+It also provides functions mimicking the `purrr` API, such as `mirai_map()`:
+
+```{r}
+
+dat |>
+  as.list(as_series = FALSE) |>
+  mirai::mirai_map(
+    \(s) {
+      stats::rnorm(s)
+    }
+  ) |>
+  _[] |>
+  as_polars_df()
+```
+
+<!-- TODO: AFAIK (etienne) there's no equivalent to `purrr::map_at()`? -->
+
+<!-- Using `mirai` directly also allows more advanced usage, such as executing a query on a remote machine. This is not in the scope of this vignette and is addressed in [vignette to write](). -->
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,8 +28,12 @@ Suggests:
     ggplot2,
     hms,
     jsonlite,
+    knitr,
+    mirai,
     nanoarrow (>= 0.6.0),
     patrick (>= 0.3.0),
+    purrr,
+    rmarkdown,
     testthat (>= 3.0.0),
     tibble,
     vctrs,
@@ -45,3 +49,4 @@ Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.2
 SystemRequirements: Cargo (Rust's package manager), rustc
+VignetteBuilder: knitr


vignettes/custom-functions.Rmd
```diff
@@ -0,0 +1,109 @@
...
+When this is the case, one should keep in mind that applying functions that are external to `polars` will limit `polars`' strengths.
+In particular, `polars` is fast in part because it is able to detect the various transformations applied to the data and run them in parallel or cache them so that only the strict minimum of computations is made.
+Using `polars` expressions will also allow the code to be run in streaming mode, allowing for larger-than-memory data.
+When one uses functions that `polars` doesn't natively know, those benefits are lost.
+This is the reason why it is better to use `polars` expressions in your own functions if you can.
+
+To apply an external function to a `polars` object, we must convert this object to R, apply the function, and then convert it back to `polars`.
+If you come from Python Polars, this is a big difference.
+In Python Polars, several functions (for instance `map_batches()` and `map_elements()`) allow one to pass custom functions.
+Those are not in the R implementation because we cannot guarantee that they will function correctly in streaming mode.
```
etiennebacher, line 57: I think this is worth mentioning, but I'm not sure we should explain why it's not in the R implementation (maybe I also missed some reasons for why it is absent).

vignettes/custom-functions.Rmd
```diff
@@ -0,0 +1,109 @@
...
+```{r}
+dat |>
+  as.list(as_series = FALSE) |>
+  purrr::map_at(c("carb", "mpg"), \(s) scale(s)[, 1], .parallel = TRUE) |>
+  _[] |>
+  as_polars_df()
+```
+
+
+## Using `mirai`
```
etiennebacher, line 88: This section is a bit empty, but I don't know what I can add. I added a comment below about running LazyFrame on a remote machine, but I have no idea how that would work (or if it even does).

If we do talk about this feature, maybe we should mention this is not the same as Polars Cloud.

etiennebacher:
@eitsupi now ready for review

etiennebacher:
Thanks for the feedback, I have addressed the version check and formatting of pkg names.

> (Considering CRAN's requirements), I do not think it is a good idea to set the number of daemons to a number greater than 1.

That would be an issue if we planned to submit to CRAN but I don't think this is on the roadmap given how upstream polars uses nightly Rust. So I don't think we're bound by CRAN settings for this.

Also, it would be confusing to users to have a section about how one can parallelize an operation but only do this on a single daemon, which seems to defeat the purpose of the parallelization. We could mention that we set it to 1 because of CRAN reasons, but that just makes the vignette more complicated for the average user.

> If you specify as_series = FALSE, I think nothing Polars specific will be lost since only the usual R vector is included.

Not sure which part of the vignette you're referring to. When I mention in the vignette that we lose some polars features, I mean the query optimizations run by polars that we lose when we convert the DataFrame to R, run some operations in parallel, and convert back to polars.

eitsupi:
> Not sure which part of the vignette you're referring to. When I mention in the vignette that we lose some polars features, I mean that we lose the query optimizations run by polars when we convert the DataFrame to R, run some operations in parallel, and convert back to polars.

I want to emphasize that there is a difference in meaning between R-vectoring a Series and then sending it to the daemon and sending the Series to the daemon and then R-vectoring it.

etiennebacher:
> I want to emphasize that there is a difference in meaning between R-vectoring a Series and then sending it to the daemon and sending the Series to the daemon and then R-vectoring it.

To me, the main difference is that converting from polars to R and then sending to the daemon is more expensive because all the conversion is done in single-threaded R, while if you send the Series to the daemon and then convert in each daemon then you parallelize this operation. Are there more advantages?

I adapted the current example to a situation where the deserialization happens in the daemon but it seems it is slower (in this particular example at least)?

``` r
library(purrr)
library(neopolars)

dat <- mtcars[, c("carb", "mpg", "drat")]
dat_all <- data.table::rbindlist(rep(list(dat), 10000)) |>
  as_polars_df()

print(dim(dat_all))
#> [1] 320000      3

# Set the number of background processes to use
mirai::daemons(3)
#> [1] 3

bench::mark(
  convert_inside_map = {
    dat_all |>
      as.list(as_series = TRUE) |>
      map(in_parallel(\(s) {
        s <- s$to_r_vector()
        scale(s)[, 1]
      })) |>
      _[] |>
      as_polars_df()
  },
  convert_outside_map = {
    dat_all |>
      as.list(as_series = FALSE) |>
      map(in_parallel(\(s) {
        scale(s)[, 1]
      })) |>
      _[] |>
      as_polars_df()
  }
)
#> # A tibble: 2 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 convert_inside_map    95.5ms   95.5ms      10.5    22.9MB     73.3
#> 2 convert_outside_map   71.6ms   71.6ms      14.0    27.4MB    112.
```

eitsupi:
My point was that there is integration with mirai regarding serialization and deserialization, so users don't have to worry about sending Series, etc. to the daemon.
If we don't mention it, it is just a common use of mirai and purrr.

But as you pointed out, it may be slower to cause the overhead of starting Polars on the daemon, since the conversion between R and Polars is performed anyway if you want to perform a function on an R vector.

But in terms of speed, I think this level of processing would hardly be faster even if it were parallelized in the first place...

``` r
library(purrr)
library(neopolars)

dat <- mtcars[, c("carb", "mpg", "drat")]
dat_all <- data.table::rbindlist(rep(list(dat), 10000)) |>
  as_polars_df()

print(dim(dat_all))
#> [1] 320000      3

mirai::daemons(0)
#> [1] 0
mirai::daemons(3)
#> [1] 3

bench::mark(
  not_parallel = {
    dat_all |>
      as.list(as_series = FALSE) |>
      map(in_parallel(\(s) {
        scale(s)[, 1]
      })) |>
      _[] |>
      as_polars_df()
  },
  parallel = {
    dat_all |>
      as.list(as_series = FALSE) |>
      map(\(s) {
        scale(s)[, 1]
      }) |>
      _[] |>
      as_polars_df()
  }
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 not_parallel     60ms   97.1ms      9.36    25.8MB     9.36
#> 2 parallel       50.8ms  121.3ms      9.77   114.2MB    39.1
```

<sup>Created on 2025-06-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

etiennebacher:
> My point was that there is integration with mirai regarding serialization and deserialization, so users don't have to worry about sending Series, etc. to the daemon.

It seems that just passing a list of Series doesn't work?

``` r
library(purrr)
library(neopolars)

dat <- as_polars_df(mtcars)

mirai::daemons(0)
#> [1] 0
mirai::daemons(3)
#> [1] 3

dat |>
  as.list(as_series = TRUE) |>
   map(in_parallel(\(s) {
     scale(s)[, 1]
   })) |>
   _[] |>
   as_polars_df()
#> Error in `map()` at neo-r-polars/R/as_polars_df.R:79:3:
#> ℹ In index: 1.
#> ℹ With name: mpg.
#> Caused by error in `array()`:
#> ! length of 'dimnames' [1] not equal to array extent
```

Even if I define:

```r
cfg <- mirai::serial_config(
  class = "polars_series",
  sfun = \(x) x$serialize(),
  ufunc = pl$deserialize_series
)
mirai::daemons(3, serial = cfg)
```
then it fails. 

Am I missing something?

> But in terms of speed, I think this level of processing would hardly be faster even if it were parallelized in the first place

It seems you swapped the "parallel" and "not_parallel" labels in your benchmark but I see your point since this `scale()` operation is quite fast anyway. 

Just to be sure, I added `Sys.sleep(1)` in all cases and the parallel gain is now clearer (as expected):

``` r
library(purrr)
library(neopolars)

dat <- mtcars
dat_all <- data.table::rbindlist(rep(list(dat), 10000)) |>
  as_polars_df()

print(dim(dat_all))
#> [1] 320000     11

mirai::daemons(0)
#> [1] 0
mirai::daemons(3)
#> [1] 3

bench::mark(
  parallel = {
    dat_all |>
      as.list(as_series = FALSE) |>
      map(in_parallel(\(s) {
        Sys.sleep(1)
        scale(s)[, 1]
      })) |>
      _[] |>
      as_polars_df()
  },
  parallel2 = {
    dat_all |>
    as.list(as_series = TRUE) |>
    map(in_parallel(\(s) {
      Sys.sleep(1)
      s <- s$to_r_vector()
      scale(s)[, 1]
    })) |>
    _[] |>
    as_polars_df()
  },
  not_parallel = {
    dat_all |>
      as.list(as_series = FALSE) |>
      map(\(s) {
        Sys.sleep(1)
        scale(s)[, 1]
      }) |>
      _[] |>
      as_polars_df()
  }
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 3 × 6
#>   expression        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 parallel         4.2s     4.2s    0.238     87.1MB    0    
#> 2 parallel2       4.23s    4.23s    0.237     81.4MB    0    
#> 3 not_parallel   11.32s   11.32s    0.0883   407.5MB    0.707
```

eitsupi:
> It seems that just passing a list of Series doesn't work?

Isn't it because you are passing a Series to the `scale` function?

vignettes/custom-functions.Rmd
```diff
@@ -0,0 +1,101 @@
...
+This is the reason why it is better to use polars expressions in your own functions if you can.
+
+To apply an external function to a polars object, we must convert this object to R, apply the function, and then convert it back to polars.
+If you come from Python Polars, this is a big difference.
+In Python Polars, several functions (for instance `map_batches()` and `map_elements()`) allow one to pass custom functions.
+Those are not in the R implementation because we cannot guarantee that they will function correctly in streaming mode.
+
+Note that when you convert your polars object to R, you may run out of memory if your object is a LazyFrame that is too big to fit in memory once collected.
+Another crucial point is that we want the R function to run as fast as possible on the data.
+One way to do that is to take advantage of purrr to run the function [in parallel](https://purrr.tidyverse.org/dev/reference/in_parallel.html).
```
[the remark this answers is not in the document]
etiennebacher: I think most people would want to click on the link instead of running an R command.

Keeping this comment open because I need to remove the "dev" from the URL once purrr 1.1.0 is on CRAN.

vignettes/custom-functions.Rmd
```diff
@@ -0,0 +1,101 @@
...
+**Note:** at the time of writing, this requires the development version of purrr (1.0.4.9000) and the package mirai.
+
+```{r, eval=check_purrr_version}
+library(purrr)
+
+# Set the number of background processes to use
+mirai::daemons(3)
+
+dat |>
+  as.list(as_series = FALSE) |>
```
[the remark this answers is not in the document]
etiennebacher: I think I'm missing something about the automatic registration (as mentioned in the [first half of this comment](https://github.com/eitsupi/neo-r-polars/pull/321#issuecomment-2994269091)) so it would be great if you could suggest the changes you would make.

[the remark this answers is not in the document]
eitsupi: After thinking again, if we do not need Series, it is sufficient to use `as.data.frame()` or `as_tibble()`.
Using `as.list()` is just confusing.

vignettes/custom-functions.Rmd
```diff
@@ -0,0 +1,99 @@
...
+
+Using custom functions is sometimes necessary when processing data.
+polars allows one to chain many expressions, thus making it possible to create custom functions that benefit from all its optimizations, such as parallelism and streaming mode.
+
+While using polars expressions is the recommended way to write custom functions, one can also apply functions by converting the DataFrame or LazyFrame to R first, and then use purrr to run functions in parallel.
+
+<!-- Reset daemons to avoid errors in other vignettes that might use them -->
+```{r, include=FALSE}
+mirai::daemons(0)
+```
```
etiennebacher, line 99: This should be visible to the users (same in the "Getting started" vignette).

eitsupi (approved):

vignettes/custom-functions.Rmd
```diff
@@ -55,45 +56,91 @@ This is the reason why it is better to use polars expressions in your own functi
...
+and only convert some Series to R, thus avoiding unnecessary conversions between
+Polars and R.
 
-```{r, eval=check_purrr_version}
+In technical terms, polars Series and other objects are wrappers around external pointers,
+which usually cannot be sent to other processes.
+However, this package has built-in integration with `{mirai}`,
+which registers the serialization and deserialization functions for Series and other objects
+when the package is loaded.
+This allows users to treat Series and other polars objects like regular R objects.
```
etiennebacher, line 121: This should be a footnote, it's not necessary to know for average users and breaks the reading flow.

etiennebacher (commented):
LGTM, thanks for finishing this. Maybe just move a paragraph to  a footnote?

vignettes/custom-functions.Rmd
eitsupi: I don't know if it is not a need-to-know for the average users, but I agree that it is worth moving to footnotes.

eitsupi: @etiennebacher altdoc with mkdocs does not seem supporting footnotes?
image

etiennebacher: I'll take a look

Issue #328, opened 2025-04-25
Do we want `[.polars_data_frame`?


etiennebacher:
The current `[.RPolarsDataFrame` is quite convoluted in the current `polars`, do we want to have this S3 method in `neopolars` as well?

https://github.com/pola-rs/r-polars/blob/a37bc3422d1a0d334b068cdf137b5ee77e210163/R/s3-methods.R#L36-L129

I think it would be nice to have but I just want confirmation before spending time on this.

eitsupi:
This would be useful because I think it could work with polars DataFrame without having to rewrite the program for data.frame.

However, after reading this document, I think it would be better to choose a behavior that matches tibble as much as possible.
<https://tibble.tidyverse.org/articles/invariants.html#definition-of-xi>

etiennebacher:
Interesting, I never read this doc. I'll take care of this one.

eitsupi:
I too did not know about this when this method was added to the current r-polars.

I think the biggest difference is whether the `drop` argument defaults to `FALSE` or `TRUE`.
`arrow::Table` also defaults to `drop = FALSE`.

etiennebacher:
For LazyFrames, the current implementation in r-polars is to error for row selection and for `drop = TRUE` with a single column.

Erroring for row selection is probably because we can't guarantee the same row ordering when repeating the same code, but why did we disallow `drop = TRUE`?

 Should we keep it like this?

eitsupi:
I don't think the process of locating rows in a LazyFrame is very expensive and should not be allowed since it expands all the data in RAM.

I think `drop` should work.

etiennebacher:
I think `[` on a LazyFrame should also return a LazyFrame and not implicitly collect. From the user perspective, it's clear that `as.data.frame()` collects the result, but maybe I just want to select columns with `[` so why should that collect the data?

If we don't collect, then we can't use the number of rows (`$height`) which is used in several places in `[.polars_data_frame`.

eitsupi:
Indeed!
In that case should the `drop` be ignored with warning or just be ignored?

etiennebacher:
We could error when `i` is not null and warn when `drop = TRUE`.

PR #330, opened 2025-04-26
feat: Add s3 method of `[` (extract) for DataFrame

etiennebacher:
Part of #328 

It quickly becomes complicated, looking at the tests first might help.


tests/testthat/test-dataframe-s3-base.R
```diff
@@ -25,3 +25,122 @@ test_that("tail() works", {
...
+  )
+  expect_identical(
+    test[1:2, ],
+    pl$DataFrame(a = 1:2, b = 4:5, c = 7:8)
+  )
+  # TODO
+  # expect_identical(
+  #   test[c(1, 10), ],
+  #   pl$DataFrame(a = c(1L, NA_integer_), b = c(4L, NA_integer_), c = c(7L, NA_integer_))
+  # )
```
etiennebacher, line 93: I didn't find a good way to include NA-filled rows for out of bound indices. There a few TODOs here related to that.

R/dataframe-s3-base.R
```diff
@@ -121,3 +121,209 @@ head.polars_data_frame <- function(x, n = 6L, ...) x$head(n = n)
...
+              x$height,
+              length(i)
+            )
+          )
+        )
+      }
+    } else {
+      # Negative indices -> drop those rows
+      # Do not accept mixing negative and positive indices
+      if (all(i < 0)) {
```
eitsupi, line 209: Is this operation safe? (i.e., does `i < 0` always return a boolean value, or can it be `NA` or non-numeric?)

eitsupi (changes requested):
Great, thanks!

R/dataframe-s3-base.R
```diff
@@ -121,3 +121,245 @@ head.polars_data_frame <- function(x, n = 6L, ...) x$head(n = n)
...
+    if (length(to_select) > 1) {
+      drop <- FALSE
+    }
+
+    x <- x$select(to_select)
+  }
+
+  if (isTRUE(drop)) {
+    as.list(x)
+  } else {
```
eitsupi, line 362: Is this correct?
I was thinking that if `drop=TRUE`, it should return Series if there is only one column, otherwise it should do nothing.

R/dataframe-s3-base.R
```diff
@@ -121,3 +121,245 @@ head.polars_data_frame <- function(x, n = 6L, ...) x$head(n = n)
...
+  }
+
+  if (is_null(i) && is_null(j)) {
+    return(x)
+  }
+
+  #### Rows -----------------------------------------------------
+
+  # check accepted types for subsetting rows
+  if (!is_null(i) && !is_character(i) && !is.numeric(i) && !is_logical(i)) {
```
eitsupi, line 173: After a little more investigation, it may be appropriate to use rlang's `is_bare_` functions here. (i.e., because S3 class vectors should not be used here)

R/dataframe-s3-base.R
etiennebacher: Actually, when we handle columns just above, we set `drop <- FALSE` if there's more than one column selected. This is also tested with `test[, c("a", "b"), drop = TRUE]` (but it's easy to miss that in the list of tests).

etiennebacher: Ah ok but we returned a list containing a single Series instead of just returning the Series, I fixed that.

eitsupi (approved):
Thank you. Let's try to merge.

PR #338, opened 2025-04-30
ci: Additional checks for `mkdocs.yml`

etiennebacher:
We already check that all files in 'man' folder are listed in mkdocs.yml. 

This PR adds two more checks:
* check that all man pages in mkdocs.yml also exist in 'man' folder. In https://github.com/eitsupi/neo-r-polars/pull/334, I didn't add the man pages in mkdocs.yml and realized afterwards that they were added there in a previous PR but were not documented in the 'man' folder.
* ~check that we don't have duplicated man pages in mkdocs.yml (`lazyframe__quantile` appeared twice)~ Could'nt make this one work. It runs fine locally but not in github actions. You can see a bunch of tries in the  "fix" commits.


eitsupi (approved):
Thanks!

etiennebacher:
I think installing R for that would be overkill. I like that this job passes or fails in 5 seconds and the error messages clearly say what's wrong when there's a failure. I don't know `yq` so I don't know its potential benefits. Feel free to rewrite those workflows but I'm not planning on spending more time on this for now.

PR #341, opened 2025-04-30
feat: Add s3 method of `[` (extract) for LazyFrame

### LINKED ISSUE #328: Do we want `[.polars_data_frame`?

etiennebacher:
Related to #328


eitsupi:
I am wondering if it is possible to rewrite the code to reduce duplicates as much as possible.

For example, this contains code older than #335 or #337, and I think that one correction should be sufficient rather than always correcting two places.

R/dataframe-s3-base.R
```diff
@@ -151,7 +160,7 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     if (!missing(drop)) {
       warn(
         c(
-          `!` = "`drop` argument ignored for subsetting a frame with `x[j]`.",
+          `!` = "`drop` argument ignored for subsetting a DataFrame with `x[j]`.",
```
etiennebacher, line 163: This is never hit when `x` is a LazyFrame because in that case `drop` is missing when we dispatched to `[.polars_data_frame`.

R/dataframe-s3-base.R
```diff
@@ -324,11 +341,18 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
...
         )
       }
       cols[j]
     } else if (is_bare_character(j)) {
+      if (!all(j %in% cols)) {
+        abort(
+          sprintf("Column(s) not found: %s.", oxford_comma(setdiff(j, cols))),
+          call = error_env
+        )
+      }
```
etiennebacher, line 355: When `x` was only a DataFrame, we didn't need this because `$select()` at the end would give a nice error message. But since `x` can now be a LazyFrame, `$select()` doesn't resolve the schema and doesn't error, so we need to "manually error" early.

etiennebacher (commented):
> I am wondering if it is possible to rewrite the code to reduce duplicates as much as possible.

R/lazyframe-s3-base.R
```diff
@@ -27,3 +27,41 @@ head.polars_lazy_frame <- head.polars_data_frame
...
+# Try to match `tibble` behavior as much as possible, following
+# https://tibble.tidyverse.org/articles/invariants.html#column-subsetting
+# TODO: add document
+#' @export
+`[.polars_lazy_frame` <- function(x, i, j, ..., drop = FALSE) {
+  # get from `[.polars_data_frame`
+  i_arg <- substitute(i)
+  # nolint start: object_usage_linter
+  j_arg <- substitute(j)
+  # nolint end
```
etiennebacher, line 41:
```suggestion
  j_arg <- substitute(j) # nolint: object_usage_linter
```

R/dataframe-s3-base.R
eitsupi: In my opinion it is needed for DataFrame too.
Since `$select()` accepts special strings like `"*"`, I think this check is necessary even if it is a DataFrame.

I will do that in #343

etiennebacher: This change applies to both DataFrame and LazyFrame in the current PR. I don't think you need to add it in #343.

R/lazyframe-s3-base.R
```diff
@@ -27,3 +27,39 @@ head.polars_lazy_frame <- head.polars_data_frame
...
+  # get from `[.polars_data_frame`
+  i_arg <- substitute(i)
+  j_arg <- substitute(j) # nolint: object_usage_linter
+
+  if (isTRUE(drop)) {
+    warn(c(`!` = "`drop = TRUE` is not supported for LazyFrame."))
+  }
+  if (!missing(i)) {
+    n_real_args <- nargs() - !missing(drop)
+    if (n_real_args > 2) {
```
eitsupi, line 46: Technically, I believe that if `i` is `NULL`, `TRUE`, or `FALSE`, it needs to be allowed. (Related to #343)

R/lazyframe-s3-base.R
```diff
@@ -27,3 +27,39 @@ head.polars_lazy_frame <- head.polars_data_frame
...
+      abort(
+        c(
+          `!` = "Cannot subset rows of a LazyFrame with `[`.",
+          i = "There are several functions that can be used to get a specific range of rows.",
+          `*` = "`$slice()` can be used to get a slice of rows.",
+          `*` = "`$gather_every()` can be used to take every nth row.",
+          `*` = "`$reverse()` can be used to reverse the order of rows."
+        )
```
eitsupi, line 54: Technically, I believe that if `i` is `NULL`, `TRUE`, or `FALSE`, it needs to be allowed. (Related to #343)

etiennebacher: done

etiennebacher:
Both test failures are unrelated.

eitsupi (approved):
Thanks.

The logic to do the discrimination in the calling function was something I didn't know about.
Great job!

PR #342, opened 2025-05-01
feat!: Bump polars to py-1.29.0

eitsupi:


R/expr-expr.R
```diff
@@ -2741,20 +2741,30 @@ expr__rolling_quantile <- function(
 #' Apply a rolling skew over values
 #'
 #' @inherit expr__rolling_max description params details
-#' @inheritParams expr__skew
-#'
 #' @inherit as_polars_expr return
+#' @inheritParams expr__skew
+#' @param min_samples The number of values in the window thatshould be non-null before computing
```
etiennebacher, line 2746:
```suggestion
#' @param min_samples The number of values in the window that should be non-null before computing
```

src/rust/src/conversion/mod.rs
```diff
@@ -728,6 +728,19 @@ impl TryFrom<&str> for Wrap<AsofStrategy> {
...
 }
 
+impl TryFrom<&str> for Wrap<RoundMode> {
+    type Error = savvy::Error;
+
+    fn try_from(round_mode: &str) -> Result<Self, Self::Error> {
+        let parsed = match round_mode {
+            "half_to_even" => RoundMode::HalfToEven,
+            "half_away_from_zero" => RoundMode::HalfAwayFromZero,
+            _ => return Err("unreachable".to_string().into()),
```
etiennebacher, line 738:
```suggestion
            _ => return Err("unreachable".to_string()),
```

tests/testthat/_snaps/input-csv-functions.md
```diff
@@ -72,17 +72,7 @@
...
-      The current offset in the file is 4 bytes.
-      
-      You might want to try:
-      - increasing `infer_schema_length` (e.g. `infer_schema_length=10000`),
-      - specifying correct dtype with the `schema_overrides` argument
-      - setting `ignore_errors` to `True`,
-      - adding `a` to the `null_values` list.
-      
-      Original error: ```remaining bytes non-empty```
+      ! provided schema does not match number of columns in file (2 != 3 in file)
```
etiennebacher, line 75: Is this a bug upstream or from us?

src/rust/src/conversion/mod.rs
eitsupi: This change broke the build.

etiennebacher: Shouldn't the output be `Result<Self, String>` like the rest of conversion functions? That's why it broke the build.

eitsupi: Error types are not unified. I think we need to unify everything to `savvy::Error` eventually.

eitsupi:
I think the Build WASM workflow failure is caused by the outdated emsdk (r-wasm/webr#532).

src/rust/src/conversion/mod.rs
etiennebacher: Ok but then I'd say let's do all at once in another PR then and use the same output types as before here

eitsupi: Again, there are already functions that use `savvy::Error`. There is no uniformity in which one is used.

PR #343, opened 2025-05-01
fix!: Extract (`[`) for DataFrame handles `NULL`, `FALSE` and column names correctly

eitsupi:
Part of #328

A follow up for #330

### DIFF
--- a/R/dataframe-s3-base.R
+++ b/R/dataframe-s3-base.R
@@ -326,14 +326,10 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
     if (length(non_existent_cols) > 0L) {
       abort(
         c(
-          `!` = "Can't subset columns that don't exist.",
+          "Can't subset columns that don't exist.",
           x = sprintf(
             "Columns %s don't exist.",
             oxford_comma(sprintf("`%s`", non_existent_cols), final = "and")
-          ),
-          i = sprintf(
-            "Available columns are %s.",
-            oxford_comma(sprintf("`%s`", cols), final = "and")
           )
         )
       )
--- a/R/dataframe-s3-base.R
+++ b/R/dataframe-s3-base.R
@@ -126,26 +126,23 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
 # https://tibble.tidyverse.org/articles/invariants.html#column-subsetting
 # TODO: add document
 #' @export
-`[.polars_data_frame` <- function(x, i, j, drop = FALSE) {
+`[.polars_data_frame` <- function(x, i, j, ..., drop = FALSE) {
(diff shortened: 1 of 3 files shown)


R/dataframe-s3-base.R
```diff
@@ -126,26 +126,23 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
 # https://tibble.tidyverse.org/articles/invariants.html#column-subsetting
 # TODO: add document
 #' @export
-`[.polars_data_frame` <- function(x, i, j, drop = FALSE) {
+`[.polars_data_frame` <- function(x, i, j, ..., drop = FALSE) {
```
etiennebacher, line 129: Why do you add dots? `[.tbl_df` adds some at the end but says in the docs they are ignored, and `[.data.frame` doesn't have any. I also don't see in the test which cases are enabled by this.

eitsupi: Looking at the Extract documentation, the default Extract requires that the `drop` be a named argument.
Perhaps many users are not even aware of the existence of this argument and may be confused if it is positionable.

R/dataframe-s3-base.R
```diff
@@ -286,74 +282,85 @@ tail.polars_data_frame <- function(x, n = 6L, ...) x$tail(n = n)
...
+        c(
+          `!` = "Can't subset columns that don't exist.",
+          x = sprintf(
+            "Columns %s don't exist.",
+            oxford_comma(sprintf("`%s`", non_existent_cols), final = "and")
+          ),
+          i = sprintf(
+            "Available columns are %s.",
+            oxford_comma(sprintf("`%s`", cols), final = "and")
+          )
```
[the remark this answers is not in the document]
eitsupi: I understand. Certainly tibble does not print this (I think removing it would result in the same error message as tibble altogether).

tests/testthat/test-dataframe-s3-base.R
```diff
@@ -169,4 +172,10 @@ test_that("`[`'s drop argument works correctly", {
     test[1, character(), drop = TRUE],
     as_polars_df(NULL)
   )
+
+  # drop should be named
+  expect_equal(
+    test[, "a", TRUE],
+    pl$DataFrame(a = 1:3)
+  )
```
etiennebacher, line 180: Shouldn't we error here by checking empty dots?

eitsupi: In general, I think the dots for extensions are silently ignored in S3 methods (because of the possibility of unknown arguments being passed for other classes).

etiennebacher (approved):
Ok, thanks I'll rebase #341 on `main` with this change later today.

eitsupi:
Thanks for your review!

PR #347, opened 2025-05-03
fix: Extract (`[`) for DataFrame handles `0` and empty numeric correctly

eitsupi:


tests/testthat/test-dataframe-s3-base.R
```diff
@@ -129,6 +150,10 @@ test_that("`[` operator works to subset rows only", {
     test["1", ],
     pl$DataFrame(a = 1L, b = 4L, c = 7L)
   )
+  expect_identical(
+    test["0", ],
```
eitsupi, line 154: tibble has a bug for this case (tidyverse/tibble#1636).

PR #363, opened 2025-05-30
feat: Register `mirai` serialization config automatically

eitsupi:
Related to #301

Thanks to r-lib/mirai#280, we no longer does not need to register mirai serializing config by hand.

``` r
library(neopolars)

mirai::daemons(4)
#> [1] 4

pl$DataFrame(a = 1, b = 2, c = 3, d = 4)$cast(pl$Int128) |>
  as.list() |>
  mirai::mirai_map(\(s) {
    s * 2L
  }) |>
  _[] |>
  as_polars_df()
#> shape: (1, 4)
#> ┌──────┬──────┬──────┬──────┐
#> │ a    ┆ b    ┆ c    ┆ d    │
#> │ ---  ┆ ---  ┆ ---  ┆ ---  │
#> │ i128 ┆ i128 ┆ i128 ┆ i128 │
#> ╞══════╪══════╪══════╪══════╡
#> │ 2    ┆ 4    ┆ 6    ┆ 8    │
#> └──────┴──────┴──────┴──────┘
```

<sup>Created on 2025-05-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/integration-mirai.R
+++ b/R/integration-mirai.R
@@ -1,0 +1,18 @@
+# Taken from the glue package's knitr integration
+# https://github.com/tidyverse/glue/blob/a3f80d678274ef634c10c2cb094c939b1543222a/R/zzz.R#L4-L11
+
+register_mirai_serial <- function() {
+  mirai::register_serial(
+    c("polars_data_frame", "polars_lazy_frame", "polars_series"),
+    sfunc = list(\(x) x$serialize(), \(x) x$serialize(), \(x) x$serialize()),
+    ufunc = list(pl$deserialize_df, pl$deserialize_lf, pl$deserialize_series)
+  )
+}
+
+on_load({
+  if (isNamespaceLoaded("mirai")) {
+    register_mirai_serial()
+  } else {
+    setHook(packageEvent("mirai", "onLoad"), function(...) register_mirai_serial())
+  }
+})
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,6 +28,7 @@ Suggests:
     ggplot2,
     hms,
     jsonlite,
+    mirai (>= 2.3.0),
     nanoarrow (>= 0.6.0),
     patrick (>= 0.3.0),
     testthat (>= 3.0.0),


shikokuchuo:
What's the issue you run into? Perhaps use something like:

```r
with(daemons(1), {
  test_that(...)
})
Sys.sleep(1) # possibly needed if at very end of tests, to ensure all background processes exit cleanly
```

eitsupi:
Thanks for taking a look at this!
One issue is that currently on CI we sometimes use `pkgload::load_all()` to run tests without installing packages.
https://github.com/eitsupi/neo-r-polars/blob/f962bc441e4e6a9e1ef3e8c679d1a900dccf96c4/.github/workflows/test-r.yml#L180-L181

I will check with another PR to see if this can be fixed (testthat didn't work well in my previous attempts with it installed).

Another question is that I don't know if something like `daemon(1)` will work even if the number of available threads on the machine running the test is small.

shikokuchuo:
It doesn't work if you add pkgload::load_all() within your mirai() call?

You can also use daemons(1, cleanup = FALSE) if you're doing several mirai calls.

R/integration-mirai.R
```diff
@@ -0,0 +1,20 @@
+# Taken from the glue package's knitr integration
+# https://github.com/tidyverse/glue/blob/a3f80d678274ef634c10c2cb094c939b1543222a/R/zzz.R#L4-L11
+
+register_mirai_serial <- function() {
+  if (is_installed("mirai", version = "2.3.0")) {
```
eitsupi, line 5: It seems that `requireNamespace("mirai", quietly = TRUE) && exists("register_serial", where = asNamespace("mirai"), mode = "function")` is better.

``` r
library(rlang)

bench::mark(
  is_installed = is_installed("mirai"),
  check_installed = check_installed("mirai"),
  is_installed_with_version = is_installed("mirai", version = "2.3.0"),
  check_installed_with_version = check_installed("mirai", version = "2.3.0"),
  require_namespace = requireNamespace("mirai", quietly = TRUE),
  require_namespace_with_existing = requireNamespace("mirai", quietly = TRUE) &&
    exists("register_serial", where = asNamespace("mirai"), mode = "function"),
  check = FALSE
)
#> # A tibble: 6 × 6
#>   expression                           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 is_installed                    348.93µs 397.85µs     2423.  439.69KB     16.7
#> 2 check_installed                   2.91µs    3.4µs   275310.   52.11KB     27.5
#> 3 is_installed_with_version       711.32µs 802.03µs     1226.    2.49KB     12.6
#> 4 check_installed_with_version    706.14µs 796.38µs     1234.   12.57KB     12.5
#> 5 require_namespace               662.99ns 806.99ns  1005199.        0B    101.
#> 6 require_namespace_with_existing   2.83µs   3.27µs   289754.    6.31KB      0
```

<sup>Created on 2025-05-31 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

(`is_installed` is another problem with a slow one, so I will create a PR upstream. ref: r-lib/rlang#1780)

eitsupi:
> It doesn't work if you add pkgload::load_all() within your mirai() call?

Thanks for the advice!
I tried the approach in 1798bd262eb91a4a1dfd486e4819262b6ae3cd53

PR #367, opened 2025-06-07
docs: Add vignette on performance

etiennebacher:
We already had this vignette, but I updated and re-worded some sections.

I think it's good to have this vignette so that we can explain somewhere the idea of  query optimizations, which IMO  doesn't fit in the "Get started" vignette (which is mostly there to explain the basic syntax of the package).


eitsupi (approved):
Thanks!

PR #381, opened 2025-06-09
test: Use the streaming engine for lazy query tests

eitsupi:


etiennebacher:
Maybe this should be an envvar so that we can have some CI that runs on the streaming engine and some that runs on the in-memory engine?

eitsupi:
I think it would be good to test both streaming engine and in-memory engine.
Anyway, I'm closing this because it doesn't seem like it's possible to do this right now.

Looking at Python Polars, it seems to use a Rust function that compares two DataFrames ignoring row order, so I think it could be done in the same way, but it's not a high priority.

PR #386, opened 2025-06-14
refactor: Simplify `as_polars_df.list()`

etiennebacher:
Following https://github.com/eitsupi/neo-r-polars/pull/384#discussion_r2145257484, this removes the use of `suppressWarnings()` and `is.infinite()`.

PR #387, opened 2025-06-15
refactor: Simplify `as_polars_df(<list>)`

eitsupi:
Another approach for #386

### DIFF
--- a/R/as_polars_df.R
+++ b/R/as_polars_df.R
@@ -180,17 +180,15 @@ as_polars_df.list <- function(x, ...) {
   list_of_series <- lapply(x, \(column) eval(call2("as_polars_series", column, !!!.args)))
 
   # Series with length 1 should be recycled
-  length_of_series <- vapply(list_of_series, length, integer(1))
+  unique_lengths <- vapply(list_of_series, length, integer(1)) |>
+    unique()
+  n_lengths <- length(unique_lengths)
 
-  # n_rows must be -Inf, 0, 2, 3, ..., because series with length 1 will be recycled even if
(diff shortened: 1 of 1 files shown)


R/as_polars_df.R
```diff
@@ -180,17 +180,15 @@ as_polars_df.list <- function(x, ...) {
...
-  # the other series have length 0.
-  n_rows <- suppressWarnings(max(length_of_series[length_of_series != 1L]))
-
-  list_of_plr_series <- if (is.infinite(n_rows)) {
-    # This case all series have length 1
+  list_of_plr_series <- if (n_lengths <= 1L) {
     list_of_series |>
       lapply(\(series) series$`_s`)
   } else {
+    n_rows <- max(setdiff(unique_lengths, 1L))
```
eitsupi, line 191: Using `vec[vec != 1L]` seems much faster than using `setdiff()`.

``` r
vec <- c(3L, 1L)

bench::mark(
  extract = max(vec[vec != 1L]),
  setdiff = max(setdiff(vec, 1L)),
  check = TRUE
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 extract       361ns  411.2ns  1754866.        0B      0
#> 2 setdiff      15.6µs   18.4µs    48455.        0B     24.2
```

<sup>Created on 2025-06-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Issue #199, opened 2025-03-12
Tracking issue: last things to do before replacing `r-polars` by `neopolars`


etiennebacher:
~R 4.5.0 is scheduled for April 11 and AFAIK it will break the current `r-polars`.~

Here's what we still need to do (@eitsupi feel free to edit the checklist):

- [x] finish porting DataFrame / LazyFrame / Expr functions and arguments. The [dashboard](https://etiennebacher.github.io/r-polars-dashboard/dashboard.html) is helpful to find gaps. I focus mostly on functions and args that exist in `r-polars` but not `neopolars`. Those who are missing in both can be added later. Note that some arguments are marked as missing but are not really, e.g. `subset` in `r-polars` is replaced by `...` in `neopolars`. 
- [x] Update README (#296)
- [x] ~find a way to support Expr functions in Series~ Done automatically, but this needs to be mentioned in the docs (#249)
- [x] finish setting up CI (#137)
- [x] implement feature flags (#197)
- [x] set profile via env var / add optimized profile (#402)
- [x] ~~Auto-completion~~ -> Moved to pola-rs/r-polars#1405
- [x] `polars_options()` (#310)
- [x] Serialize/Deserialize functions for DataFrame, LazyFrame, Series.
- [x] Documentation
  - [x] reorganize and clean up the NEWS (#284)
  - [x] add the documentation infrastructure (#254)
  - [x] add vignette(s) about custom user functions and `mirai` (#321)
  - [x] add "Get started" vignette (#279)
  - [x] add installation vignette (#413)
- [x] implement `arr$to_struct()` and `list$to_struct()` (#420, #423)
- [x] fix warnings in R CMD check (#396)

eitsupi:
Thanks for summarizing this!

I believe the current r-polars will work with R 4.5.0, but I do think it is possible that they will be removed from the R-multiverse Production repository due to R CMD check warnings or failures......

eitsupi:
Oops, sorry, I may have underestimated the situation, but looking at the R-universe, it appears that a large number of tests are failing in R 4.5.
Like:

```log
  > df = pl$DataFrame(a = c(1, 2, 3), b = c(4, 5, 6))
  > 
  > df$select((pl$col("a") * pl$col("b"))$sum())$item()
  Error in islistfactor(x, recursive) : NA in coercion to boolean
  Calls: <Anonymous> ... as.vector -> as.vector.RPolarsSeries -> <Anonymous> -> unlist
[...]

PR #395, opened 2025-06-19
chore!: Rename the package from neopolars to polars and bump version

eitsupi:
I think we are ready to rename.
Context: pola-rs/r-polars#1336


etiennebacher:
Just to be sure, it seems fine to me to rename now but we should implement the two `$to_struct()` functions before the release (cf https://github.com/eitsupi/neo-r-polars/issues/151 and mentioned in https://github.com/eitsupi/neo-r-polars/issues/199)

eitsupi:
@etiennebacher The mysterious bug that was occurring on Windows has disappeared and I would like to merge this in, is that ok?

etiennebacher:
Which bug are you referring to?

I'm fine with merging this, I simply want to be sure #199 is completed before making a new release.

eitsupi:
> Which bug are you referring to?

This error.
https://github.com/eitsupi/neo-r-polars/actions/runs/15748679776/job/44454246780

```
C:\rtools43\x86_64-w64-mingw32.static.posix\bin/ld.exe: error: export ordinal too large: 267626

collect2.exe: error: ld returned 1 exit status
if [ "true" != "true" ]; then \
	rm -Rf "/d/a/neo-r-polars/neo-r-polars/src/rust/target//debug/libr_polars.a" "" "" "/d/a/neo-r-polars/neo-r-polars/src/rust/target//debug/build"; \
fi
no DLL was created

ERROR: compilation failed for package 'polars'
```

The same error occurred after re-running CI three times, so it is reproducible. I really don't know why it only occurred on this PR.

Issue #396, opened 2025-06-19
Fix various warnings in R CMD check


etiennebacher:
Just keeping track of this since warnings don't fail the CI workflow (for now)

@eitsupi you can assign this to me

PR #399, opened 2025-06-19
docs: Add documentation for all S3 arithmetic methods

etiennebacher:
Part of  #396 

Almost an exact copy-paste of r-polars docs. This should remove most (but not all) R CMD check warnings about missing links.


eitsupi (approved):
Thanks!

PR #400, opened 2025-06-21
docs: More fixes for R CMD check warnings

etiennebacher:
Part of #396 

In this PR:

* drop examples of deprecated functions since they throw a warning
* remove links to `GroupBy_class` and similar since their docs are not very useful (for instance in r-polars, `RollingGroupBy_class` only said that this class comes from `$rolling()`). Instead, I mention the class of the output
* fix other cross links

### DIFF
--- a/R/expr-categorical.R
+++ b/R/expr-categorical.R
@@ -36,6 +36,8 @@ expr_cat_get_categories <- function() {
 
 #' Set Ordering
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
 #' Determine how this categorical series should be sorted.
 #'
 #' @param ordering string either 'physical' or 'lexical'
--- a/R/expr-datetime.R
+++ b/R/expr-datetime.R
@@ -702,7 +702,8 @@ expr_dt_timestamp <- function(time_unit = c("us", "ns", "ms")) {
 # TODO: mark deprecated in news before next release
 #' Set time unit of a Series of dtype Datetime or Duration
 #' @description
-#' This is deprecated. Cast to Int64 and then to Datetime instead.
+#' `r lifecycle::badge("deprecated")`
+#' Cast to Int64 and then to Datetime instead.
 #'
 #' @inheritParams expr_dt_timestamp
 #' @inherit as_polars_expr return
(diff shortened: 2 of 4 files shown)


R/datatypes-classes.R
```diff
@@ -20,6 +20,12 @@
 #' see the section about floating point numbers for more information.
 #'
 #' @details
+#' ## Datetime representation
+#'
+#' The underlying representation of datetime is a 64-bit signed integer. The
+#' integer indicates the number of time units since the Unix epoch (1970-01-01
+#' 00:00:00). The number can be negative to indicate datetimes before the epoch.
+#'
```
etiennebacher, line 28: In old r-polars, we had a specific `DataType_Datetime` page, but here all datatypes link to this page (e.g. `?pl__Datetime`). Therefore, I added this extra paragraph which was the only thing missing from the old  `DataType_Datetime` docs.

eitsupi: Do we need this? I don't see the need for it. (I mean, why is there an explanation for internal representation only for Datetime?)

eitsupi (commented):
Thanks!
One minor comment.

R/datatypes-classes.R
eitsupi: > casting a date like "1950-01-01" to integer returns a negative integer.

I believe this is the same for the base R Date type.

``` r
as.Date("1950-01-01") |> as.integer()
#> [1] -7305
```

<sup>Created on 2025-06-22 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I don't think we should create this here, since it is a common way of representing time types to hold the elapsed time from POSIX origin as an integer.

And if something like this is needed, shouldn't it be in the table below?
Anyway I strongly object to adding this here just because it was in the old r-polars.

etiennebacher: Dropped this part in [0b51ac7](https://github.com/eitsupi/neo-r-polars/pull/400/commits/0b51ac7a418df4eefd76b2b0e96eae04e0822471)

eitsupi (commented):
Could you add the deprecation badge to the functions from which you removed the examples?

etiennebacher:
Done in [9fbc732](https://github.com/eitsupi/neo-r-polars/pull/400/commits/9fbc7324d13663c2147a9053d58afcaf70f732a1)

PR #414, opened 2025-07-05
feat: Allow `$width` and `$columns` on LazyFrame

### LINKED ISSUE #409: Allow `$width` on LazyFrame
This was allowed in r-polars, and it raises a warning but still works in python polars:

```python
import polars as pl

pl.LazyFrame({"x": [1, None, 3]}).width

# PerformanceWarning: determining the width of a LazyFrame requires
# resolving its schema, which is a potentially expensive operation. Use 
# `LazyFrame.collect_schema().len()` to get the width without this warning.
# 1
```

Mentioned in https://github.com/pola-rs/r-polars/issues/1400

etiennebacher:
Both are allowed and throw a warning in python polars.

Fixes #409


R/lazyframe-frame.R
```diff
@@ -9,6 +9,11 @@
 #' @aliases polars_lazy_frame LazyFrame
 #' @inheritParams pl__DataFrame
 #' @return A polars [LazyFrame]
+#'
+#' @section Active bindings:
+#' - `columns`: `$columns` returns a character vector with the names of the columns.
+#' - `width`: `$width` returns a integer with the number of columns of the DataFrame.
+#'
```
eitsupi, line 16: We do not recommend their use and therefore they should not be documented.

R/lazyframe-frame.R
```diff
@@ -35,6 +40,35 @@ wrap.PlRLazyFrame <- function(x, ...) {
...
   self$`_ldf` <- x
 
+  # nolint start: line_length_linter
+  makeActiveBinding(
+    "columns",
+    function() {
+      warn(
+        c(
+          "Determining the column names of a LazyFrame requires resolving its schema, which is a potentially expensive operation.",
+          "i" = "Use `names(<lazyframe>$collect_schema())` to get the column names without this warning."
```
eitsupi, line 50: Just `names(<lazyframe>)` is sufficient.

R/lazyframe-frame.R
```diff
@@ -35,6 +40,35 @@ wrap.PlRLazyFrame <- function(x, ...) {
...
+    },
+    self
+  )
+  makeActiveBinding(
+    "width",
+    function() {
+      warn(
+        c(
+          "Determining the width of a LazyFrame requires resolving its schema, which is a potentially expensive operation.",
+          "i" = "Use `length(<lazyframe>$collect_schema())` to get the width without this warning."
```
eitsupi, line 63: Just `length(<lazyframe>)` is sufficient.

eitsupi (changes requested):

eitsupi (approved):

PR #419, opened 2025-07-06
feat: Add `arr$to_struct()`

etiennebacher:
Related to #151 

This does **not** compile, I just save it here in case it's helpful. The error here is the following:

```
`*mut c_void` cannot be shared between threads safely
...
= help: within `{closure@src/expr/array.rs:122:22: 122:39}`, the trait `Sync` is not implemented for `*mut c_void`
note: required because it appears within the type `FunctionSexp`
   --> /home/etienne/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/savvy-0.8.10/src/sexp/function.rs:11:12
    |
11  | pub struct FunctionSexp(pub SEXP);
    |            ^^^^^^^^^^^^
note: required because it's used within this closure
   --> src/expr/array.rs:122:22
    |
122 |             Arc::new(move |idx: usize| {
    |                      ^^^^^^^^^^^^^^^^^
    = note: required for the cast from `std::sync::Arc<{closure@src/expr/array.rs:122:22: 122:39}>` to `std::sync::Arc<(dyn Fn(usize) -> polars::prelude::PlSmallStr + Send + Sync + 'static)>`
```

I don't know if this is related to https://github.com/yutannihilation/savvy/issues/154 or not.

@eitsupi do you have an idea of how to implement it?

### DIFF
--- a/src/rust/src/expr/array.rs
+++ b/src/rust/src/expr/array.rs
@@ -1,5 +1,12 @@
-use crate::{PlRExpr, prelude::*};
-use savvy::{NumericScalar, Result, savvy};
+use crate::{
+    PlRExpr,
+    error::RPolarsErr,
+    prelude::*,
+    r_threads::ThreadCom,
+    r_udf::{CONFIG, RUdf},
+};
+use polars::prelude::array::ArrToStructNameGenerator;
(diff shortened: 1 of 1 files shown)


eitsupi:
My attempt is #420.

As with map_batches, I am not actively looking to implement this because it requires relying on ThreadCom.
However, it does seem to work.

PR #420, opened 2025-07-06
feat: Add `<expr>$arr$to_struct()`

eitsupi:
I first tried it without ThreadCom, and although it compiled successfully, R crashed when I ran it.


eitsupi:
I will work on `list_to_struct` tomorrow.

PR #436, opened 2025-07-16
feat: ``<lazyframe>$`_fetch`() ``

eitsupi:


etiennebacher:
>  This difference results in very different behavior when executing large queries on large data sets.

Fair enough, we just don't use large data in the test suite so I'd rather not make the API (even internal only) heavier with this function.

eitsupi:
> we just don't use large data in the test suite so I'd rather not make the API (even internal only) heavier with this function.

I'm sorry, but I'm not sure what you are trying to say.
By "for testing" here, in my opinion, this means not for package developers, but for advanced users writing queries.

etiennebacher:
Ok, thanks for asking upstream

etiennebacher:
Out of curiosity, in what setting did you have a panic?

eitsupi:
> Out of curiosity, in what setting did you have a panic?

I noticed this this time when I ran `fetch`, but I think this was causing the session to crash in webR as well before `map_batches` was disabled on wasm32.

I believe this also occurred when `map_batches` were run in a sequence.
(I believe this is one of the reasons why `map_batches` is buggy.)

PR #466, opened 2025-07-19
fix: Update arguments of `$and()` and `$or()` to match Python Polars

etiennebacher:
Fixes https://github.com/pola-rs/r-polars/issues/1417


eitsupi (approved):
Thanks!

PR #468, opened 2025-07-20
fix: `round()` uses `decimals = 0` as default value

etiennebacher:
Also some docs improvements


eitsupi (changes requested):
Thanks!
Could you use `0L` instead of `0`?

eitsupi (approved):

